PR #19, opened 2022-07-22
feat: Add os env variable validation

long-sean:
Informs users of missing environment variables at run time: 
 - Add hashicorp/multierror
 - Add OsEnvironmentError type
 - Validation function for NewFromEnvironment

1err

2err


borland (approved):
LGTM

PR #27, opened 2022-07-25
Orion/space selector

borland:

PR #28, opened 2022-07-25
chore: survey mocking

domenicsim1:


pkg/cmd/account/list/list.go
```diff
@@ -18,7 +18,7 @@ func NewCmdList(client apiclient.ClientFactory) *cobra.Command {
 		`),
 		Aliases: []string{"ls"},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			client, err := client.Get(true)
+			client, err := client.Client(true)
```
borland, line 21: I refactored this to GetSystemClient and GetSpacedClient so there's going to be loads of merge conflicts. I'm happy to resolve them if you merge yours first

pkg/question/helpers_test.go
```diff
@@ -0,0 +1,67 @@
...
+func (qa *QA) String() string {
+	return fmt.Sprintf("{\n"+
+		"	prompt: %+v\n"+
+		"	answer: %v\n"+
+		"	err: %v"+
+		"\n}\n", qa.prompt, qa.answer, qa.err)
+}
+
+func NewAskMocker(t assert.TestingT, questions []QA) (question.Asker, func()) {
+	c := make(chan QA)
```
borland, line 27: What's the channel doing here? Do we have any actual concurrency?

It appears as though you're front-loading all the questions into the channel's internal buffer then pulling one item out each time a question gets asked. If that's the case then that appears to be equivalent to a plain old array-based queue; you could simply pop the first item off an array of questions, couldn't you?

e.g line 86 here:
https://github.com/OctopusDeploy/cli/pull/27/files#diff-969618f76259350f331567e47f400bdf86b86f0e4c5b95989df775b9195415d5R86

pkg/factory/factory.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"github.com/AlecAivazis/survey/v2"
+	"github.com/OctopusDeploy/cli/pkg/apiclient"
+	"github.com/OctopusDeploy/cli/pkg/question"
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/client"
+)
+
+type factory struct {
+	client apiclient.ClientFactory
+	asker  question.Asker
```
borland, line 12: Happy to see you've gone with a dependency injection based approach rather than a global :-)

borland (approved):

pkg/question/helpers_test.go
domenicsim1: Yes, having an array that dequeues could work here. I decided to use a generator pattern. The channel does not actually front-load, sorta. `c <- qa`  is sorta like yield, it will wait for the channel to be read before looping again. Agreed this is probably "over engineered" but It works so I will leave it as is for now.

borland: Yeah I didn't block the PR on behalf of this, but it is a bit weird; Seeing channels in the code signals "concurrency" which is a false signal here

pkg/cmd/account/list/list.go
domenicsim1: 👌

PR #40, opened 2022-07-27
Orion/map collection with lookups

borland:


pkg/cmd/release/list/list.go
```diff
@@ -0,0 +1,121 @@
...
+				return err
+			}
+
+			type ReleaseOutput struct {
+				Channel string
+				Project string
+				Version string
+			}
+
+			// page-fetching loop. TODO sync with dom
```
domenicsim1, line 38: Still want to sync?

pkg/cmd/release/list/list.go
```diff
@@ -0,0 +1,128 @@
...
+
+			type ReleaseViewModel struct {
+				Channel   string
+				ChannelID string `json:",omitempty"`
+				Project   string
+				ProjectID string `json:",omitempty"`
+				Version   string
+			}
+
+			// page-fetching loop. TODO sync with dom
```
domenicsim1, line 38: Still want to sync?

pkg/util/util.go
```diff
@@ -0,0 +1,134 @@
...
+func SliceContains[T comparable](slice []T, target T) bool {
+	for _, item := range slice {
+		if item == target {
+			return true
+		}
+	}
+	return false
+}
+
+func MapSlice[T any, TResult any](slice []T, mapper func(item T) TResult) []TResult {
```
borland, line 12: This is awkward. Every language other than C# calls this operation "map", however golang calls dictionaries "map" as well, so there's a confusion of terms.

Maybe we should rename it to `SliceSelect` to match C# or something? Perhaps `TransformSlice` to align with `std::transform` in C++ instead?. Or we could just leave it 🤷

domenicsim1: Let’s use SliceTransform

domenicsim1 (approved):
😎

pkg/util/util.go
borland: Done :-)

pkg/cmd/release/list/list.go
borland: I'll merge to main first and sync as part of create-release; that part is lower priority

PR #41, opened 2022-07-28
feat: account <type> list commands

domenicsim1:


pkg/cmd/account/list/list.go
```diff
@@ -32,17 +36,25 @@ func NewCmdList(client factory.Factory) *cobra.Command {
 				Id   string
 				Name string
 				Type string
-				// TODO should we list type-specific fields here? (e.g. an Azure DevOps account may have different attributes than a Username)
 			}
 
-			return output.PrintArray(allAccounts, cmd, output.Mappers[accounts.IAccount]{
+			accountTypeMap := map[accounts.AccountType]string{
```
slewis74, line 41: I wonder if the Go client would ultimately be a better place for this to live? Doesn't have to be for now, maybe we should consider as for v3 of the client?

domenicsim1: 100% we should consider moving this at some point.

pkg/cmd/account/aws/list/list.go
```diff
@@ -0,0 +1,72 @@
...
+		AccountType: accounts.AccountTypeAmazonWebServicesAccount,
+	})
+	if err != nil {
+		return err
+	}
+	items, err := accountResources.GetAllPages(client.Accounts.GetClient())
+	if err != nil {
+		return err
+	}
+	s.Stop()
```
borland, line 46: do you need to `defer` this? If we return early with an error will the spinner just keep running, or is there some wrapper which handles it externally?

pkg/cmd/account/aws/list/list.go
```diff
@@ -0,0 +1,72 @@
...
+
+	output.PrintArray(items, cmd, output.Mappers[accounts.IAccount]{
+		Json: func(item accounts.IAccount) any {
+			acc := item.(*accounts.AmazonWebServicesAccount)
+			return &struct {
```
borland, line 51: Cool! TIL you can have anonymous structs

pkg/cmd/account/azure/list/list.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	azureEnvMap := map[string]string{
+		"":                  "Global Cloud",
+		"AzureCloud":        "Global Cloud",
+		"AzureChinaCloud":   "China Cloud",
+		"AzureGermanCloud":  "German Cloud",
+		"AzureUSGovernment": "US Government",
+	}
+	s.Stop()
```
borland, line 54: as with AWS, do we need to defer the stop?

pkg/cmd/account/azure/list/list.go
```diff
@@ -0,0 +1,86 @@
...
+				TenantId:           acc.TenantID.String(),
+				ClientId:           acc.ApplicationID.String(),
+				AzureEnvironment:   acc.AzureEnvironment,
+			}
+		},
+		Table: output.TableDefinition[accounts.IAccount]{
+			Header: []string{"NAME", "SUBSCRIPTION ID", "TENANT ID", "APPLICATION ID", "AZURE ENV"},
+			Row: func(item accounts.IAccount) []string {
+				acc := item.(*accounts.AzureServicePrincipalAccount)
+				return []string{output.Bold(acc.GetName()), acc.SubscriptionID.String(), acc.TenantID.String(), acc.ApplicationID.String(), azureEnvMap[acc.AzureEnvironment]}
```
borland, line 79: this line is a bit long

borland:
Accounts all have a description, and optionally a set of environments that the account is restricted for use in. Should we return any of that stuff when listing with table or JSON? Personally I'd be happy to create a github issue to track it and kick it down the line for a while until after account create is done, etc (we may want to implement 'account view' or something like that to cover all the details)

borland (approved):

pkg/cmd/account/aws/list/list.go
domenicsim1: good catch with the spinner, if the program exists the spinner stops and cleans itself. So errors work fine. BUT to be safe I added some stops before the returns on the errors as it does feel safer. Sadly you can not use defer on spinner stop, it messes up leaves the spinner before the start of the printed text and the terminal in raw mode after the program exits

PR #42, opened 2022-08-01
Orion/create release

borland:
Implements the basic flow for create release where it will prompt for Project + Channel + Version or accept them from the command line


pkg/cmd/release/create/create.go
```diff
@@ -19,13 +52,320 @@ func NewCmdCreate(f factory.Factory) *cobra.Command {
...
+		projectsPage, err = projectsPage.GetNextPage(octopus.Projects.GetClient())
+		if err != nil {
+			return nil, err
+		} // if there are no more pages, then GetNextPage will return nil, which breaks us out of the loop
+	}
+
+	return nil, fmt.Errorf("no project found with name of %s", projectName)
+}
+
+func selectPackageOverrides(octopus *octopusApiClient.Client, ask question.Asker, project *projects.Project, channel *channels.Channel, releaseID string) (string, error) {
```
borland, line 250: This function is still JB's unmodified prototype code; I haven't cleaned it up, so please disregard any code review concerns you might have here. Will circle back in a future PR when the package override stuff is done properly

jbristowe:
image

I've noticed that the spinner doesn't revert back to a selector arrow (`>`) once the spaces are loaded. Any chance of fixing this in this PR?

jbristowe (approved):

borland:
@jbristowe I fixed the spinner by not starting the spinner on a call to GetSpacedClient.

we've discovered an anti-pattern here: If you start the spinner on an outer call, then somewhere nested inside that there's a prompt for user input, the prompting code doesn't neccessarily know about spinners in order to stop them.

Basically: Don't call spinner when you might do something that prompts for user input

PR #44, opened 2022-08-02
Orion/create release integration 1

borland:


cmd/octopus/main.go
```diff
@@ -22,15 +23,24 @@ func main() {
 	// if there is a missing or invalid .env file anywhere, we don't care, just ignore it
 	_ = godotenv.Load()
 
-	clientFactory, err := apiclient.NewClientFactoryFromEnvironment()
+	// initialize our wrapper around survey, which is also used as a flag for whether
+	// we are in interactive mode or automation mode
+	askProvider := question.NewAskProvider(survey.AskOne)
```
borland, line 28: @domenicsim1 this is significant. Both the ClientFactory and the proper factory need a reference to Survey, and it gets nilled out when you specify --no-prompt on the command line. I had to make a third struct to hold the Survey reference to avoid an import loop between ClientFactory <-> Factory.

Seems to work well enough

pkg/apiclient/client_factory.go
```diff
@@ -64,11 +63,29 @@ type Client struct {
...
+
+		if vstr, ok := v.(string); ok && vstr == "" {
+			return fmt.Errorf("required string argument %s was empty", k)
+		}
+	}
+	return nil
+}
+
+func NewClientFactory(httpClient *http.Client, host string, apiKey string, spaceNameOrID string, ask question.AskProvider) (ClientFactory, error) {
+	if err := requireArguments(map[string]any{
```
borland, line 83: This doesn't seem particularly "go-ish" but google searching kept sending me down the path of using big validation libraries that use reflection etc... which seemed way to heavy just to assert that some parameters aren't nil. Open to suggestions here on how we might more easily validate non-nil parameters

test/integration/release_create_test.go
```diff
@@ -23,3 +41,194 @@ package integration
...
+
+func TestReleaseCreate(t *testing.T) {
+	_, isGithubActions := os.LookupEnv("GITHUB_ACTIONS")
+	if isGithubActions {
+		// temp hack to disable these tests when running in Github actions
+		// because at time of writing, the current public octopus server docker image is version 2022.2,
+		// and the API's we are trying to call don't exist until 2022.3. The API's also don't
+		// expose hypermedia links that we can use to feature-detect, and the Go SDK library doesn't
+		// expose the server version either.
+		return
```
borland, line 65: 🤮

.github/workflows/test.yml
```diff
@@ -28,7 +28,10 @@ jobs:
           --health-retries 10
           --health-start-period 10s
       octopusserver:
-        image: octopusdeploy/octopusdeploy
+        # NOTE: we'd rather pin to a major version (e.g. 2022.3), but because 2022.3 is not
+        # generally available for onprem customers yet this isn't possible.
+        # Change this when it GA's later
+        image: docker.packages.octopushq.com/octopusdeploy/octopusdeploy:2022.3.7064-linux
```
jbristowe, line 34: Great find! 🎉

borland: 👍 @hnrkndrssn

cmd/octopus/main.go
```diff
@@ -2,12 +2,13 @@ package main
 
 import (
 	"fmt"
+	"github.com/AlecAivazis/survey/v2"
 	"github.com/OctopusDeploy/cli/pkg/constants"
+	"github.com/OctopusDeploy/cli/pkg/question"
 	"github.com/spf13/cobra"
 	"os"
 	"time"
```
jbristowe, line 10: Sorting makes me happy. 🤗

cmd/octopus/main.go
```diff
@@ -22,15 +23,24 @@ func main() {
...
 
-	clientFactory, err := apiclient.NewClientFactoryFromEnvironment()
+	// initialize our wrapper around survey, which is also used as a flag for whether
+	// we are in interactive mode or automation mode
+	askProvider := question.NewAskProvider(survey.AskOne)
+	_, ci := os.LookupEnv("CI")
+	// TODO move this to some other function and have it look for GITHUB_ACTIONS etc as we learn more about it
+	if ci {
+		askProvider.SetPromptDisabled()
+	}
```
jbristowe, line 33: I believe we'll need to create a version of the [AutomationEnvironmentProvider](https://github.com/OctopusDeploy/OctopusClients/blob/master/source/Octopus.Server.Client/AutomationEnvironments/AutomationEnvironmentProvider.cs) available in the C# API client. This should provide a more reliable detection mechanism since `CI` isn't honoured by every automation environment.

jbristowe:
I noticed that you've added the binary/build in the commit (i.e. `cmd/octopus/octopus`). This should be removed.

pkg/apiclient/client_factory.go
jbristowe: Yes, there are validation libraries to consider. I don't think we should use them in this instance. Go is very idiomatic when it comes to input validation for functions. Everyone writes the same checks:

```go
if inputParameter == nil {
  return fmt.Errorf("...", ...)
}
```

pkg/apiclient/client_factory.go
```diff
@@ -82,26 +99,18 @@ func NewClientFactory(httpClient *http.Client, host string, apiKey string, space
...
 	return clientImpl, nil
 }
 
 // NewClientFactoryFromEnvironment Creates a new Client wrapper structure by reading the environment.
 // specifies nil for the HTTP Client, so this is not for unit tests; use NewClientFactory(... instead)
-func NewClientFactoryFromEnvironment() (ClientFactory, error) {
+func NewClientFactoryFromEnvironment(ask question.AskProvider) (ClientFactory, error) {
 	host := os.Getenv("OCTOPUS_HOST")
 	apiKey := os.Getenv("OCTOPUS_API_KEY")
 	spaceNameOrID := os.Getenv("OCTOPUS_SPACE")
```
jbristowe, line 112: We should provide turn these into constants.

pkg/apiclient/client_factory.go
```diff
@@ -154,7 +163,7 @@ func (c *Client) GetSpacedClient() (*octopusApiClient.Client, error) {
 	var foundSpaceID string
 	// if c.Ask is nil it means we're in automation mode.
 	if c.SpaceNameOrID == "" {
-		if c.Ask == nil {
+		if !c.Ask.IsPromptEnabled() {
```
jbristowe, line 166: Can we change `IsPromptEnabled` to `IsInteractive` or something like that? I'd like to aim for a consistent term that we use to describe the situation where a human is using our CLI. We've called this "interactive mode" in our docs.

pkg/cmd/release/create/create.go
```diff
@@ -49,7 +48,7 @@ func NewCmdCreate(f factory.Factory) *cobra.Command {
 		Short: "Creates a release in an instance of Octopus Deploy",
 		Long:  "Creates a release in an instance of Octopus Deploy.",
 		Example: fmt.Sprintf(heredoc.Doc(`
-			$ %s release create"
+			$ %s release create --project MyProject --channel Beta -v "1.2.3"
```
jbristowe, line 51: ❤️

pkg/cmd/release/create/create.go
```diff
@@ -98,14 +95,23 @@ func createRun(f factory.Factory, w io.Writer, options *executor.TaskOptionsCrea
...
+
+	// the executor will raise errors if any required options are missing
+	err = executor.ProcessTasks(f, []*executor.Task{executor.NewTask(executor.TaskTypeCreateRelease, options)})
+	if err != nil {
+		return err
+	}
+
+	if options.Response != nil {
+		// TODO AutomaticallyDeployedEnvironments. Discuss with Team
+		_, err = fmt.Fprintf(w, "Successfully created Release Version %s %s.\n", options.Response.ReleaseVersion, output.Dimf("(%s)", options.Response.ReleaseID))
```
jbristowe, line 108: This should probably read:

```go
_, err = fmt.Fprintf(w, "Successfully created release %s with version, %s\n", output.Dimf("(%s)", options.Response.ReleaseID), options.Response.ReleaseVersion)
```

pkg/executor/release.go
```diff
@@ -15,6 +19,9 @@ type TaskOptionsCreateRelease struct {
 	ChannelName  string // optional
 	// TODO array of package version overrides
 	ReleaseNotes string // optional
+
+	// if the task succeeds, the resulting output will be stored here
+	Response *releases.CreateReleaseResponseV1
```
jbristowe, line 24: 🕺

pkg/executor/release.go
```diff
@@ -33,21 +40,30 @@ func releaseCreate(f factory.Factory, input any) error {
 
 	// we have the provided project name; go look it up
 	if params.ProjectName == "" {
-		return errors.New("Project must be specified")
+		return errors.New("project must be specified")
```
jbristowe, line 43: 👍

pkg/apiclient/client_factory.go
borland: I thought about that, but then a run of 6 `if x == nil`'s seemed like it'd be a bit long.

cmd/octopus/main.go
borland: Thanks for the pointer to the existing code, I wasn't aware of that. Agree; the TODO indicates where we'd call it here

pkg/apiclient/client_factory.go
borland: will do

pkg/apiclient/client_factory.go
borland: sounds good

jbristowe (approved):

PR #45, opened 2022-08-02
feat: create AWS account

domenicsim1:


pkg/cmd/account/aws/create/create.go
```diff
@@ -0,0 +1,173 @@
...
+			opts.Octopus = client
+			opts.Writer = cmd.OutOrStdout()
+			if descriptionFilePath != "" {
+				data, err := os.ReadFile(descriptionFilePath)
+				if err != nil {
+					return err
+				}
+				opts.Description = string(data)
+			}
+			noPrompt, err := cmd.Flags().GetBool(constants.FlagNoPrompt)
```
borland, line 67: If factory.Ask is nil, it means we're in CI mode; otherwise we're in interactive mode. The decision making (flags, environment variables, etc) is centralised down in the factory. There's an `IsPromptEnabled()` helper method on the factory that you can check for a nice-looking version of the nil check.

Suggest you use that rather than checking flags specifically here

pkg/cmd/account/helper/helper.go
```diff
@@ -0,0 +1,38 @@
...
+		if err != nil {
+			return nil, err
+		}
+		allMatches, err := matches.GetAllPages(octopus.Environments.GetClient())
+		if err != nil {
+			spinner.Stop()
+			return nil, err
+		}
+		for _, match := range allMatches {
+			if envName == match.Name {
```
borland, line 29: I used `strings.EqualFold` in create release for case-insensitivity... felt like the right move for things that people type in by hand🤷 
You may want to consider using it here.

pkg/cmd/account/helper/helper.go
```diff
@@ -0,0 +1,38 @@
...
+				envIds = append(envIds, match.ID)
+				continue loop
```
borland, line 31: I had to think pretty hard to work out the logic here...

Perhaps it would be simpler just to do an ordinary loop with a break (and no continue-outer) and then just put an if statement at the end which says `if(len(envIds) == 0) { envIds = envIds.Append(envName) }`

pkg/cmd/environment/list/list.go
```diff
@@ -28,14 +28,14 @@ func NewCmdList(f factory.Factory) *cobra.Command {
 				return err
 			}
 
-			envResources, err := client.Environments.Get(environments.EnvironmentsQuery{
-				Take: 1,
-			})
+			envResources, err := client.Environments.Get(environments.EnvironmentsQuery{})
 			if err != nil {
+				f.Spinner().Stop()
```
borland, line 33: Hmmm, this spinner is getting a bit awkward, I wonder if we can do something to help with defer etc. I have no solutions at the moment, but it's something to think about in the background

pkg/question/selectors/environments.go
```diff
@@ -0,0 +1,34 @@
...
+		return nil, err
+	}
+	s.Stop()
+	items, err := question.MultiSelectMap(ask, message, allEnvs, func(item *environments.Environment) string {
+		return item.Name
+	})
+	if err != nil {
+		return nil, err
+	}
+	itemIds := make([]string, 0, len(items))
```
borland, line 29: you could use util.SliceTransform here.

I'm unsure if we should though? How good is go at inlining lambdas?

pkg/surveyext/editor.go
```diff
@@ -0,0 +1,222 @@
+package surveyext
+
+// This file extends survey.Editor to add an "Optional" flag
+// and change the default editor to nano on unix os.
```
borland, line 4: Why this change?

I <3 vim, and presumably survey has some good reason to pick it as the default?

borland (approved):

pkg/surveyext/editor.go
domenicsim1: I <3 vim too but many people try to avoid it. The Github cli also defaults to nano.

PR #48, opened 2022-08-09
Orion/create release 2

borland:
chore: Splits unit and integration tests into two workflows. Integration runs nightly, unit runs on every push
refactor: Cover release create workflow with unit tests mocking out the octopus server
feat: release create supports --git-commit and --git-ref on the command line


pkg/cmd/release/create/create.go
```diff
@@ -135,6 +192,11 @@ func AskQuestions(octopus *octopusApiClient.Client, asker question.Asker, option
 	}
 	options.ProjectName = selectedProject.Name
 
+	if selectedProject.PersistenceSettings.GetType() == "VersionControlled" && options.GitReference != "" || options.GitCommit != "" {
+		// it's a config-as-code project, we need to ask for Git Ref or Commit
+
+	}
```
jbristowe, line 198: Yuck. We should fix this in the Go API client somehow. But I understand why you're doing it. 😄

jbristowe (approved):

pkg/cmd/release/create/create.go
borland: The `"PersistenceSettings.GetType() == "VersionControlled""` magic string bit?
That seemed to be how the go API client was doing it, so I copied from there. We could use type aliased strings like `AccountType` does? That seems to be the closest equivalent Go has to enums?

Should I file an issue against the client repo? for this

jbristowe: I'm wondering if there's a way we can hide this; put a better veneer on it.

I built this support for CaC back when it was available as an EAP. I kept them as strings in case things changed. Perhaps it's time to re-visit.

borland: OK, I'll create the issue and we can have the discussion over there

borland: Issue filed:

https://github.com/OctopusDeploy/go-octopusdeploy/issues/133

PR #51, opened 2022-08-17
Orion/create release 4

borland:


go.mod
```diff
@@ -2,10 +2,12 @@ module github.com/OctopusDeploy/cli
 
 go 1.18
```
jbristowe, line 3: We should jump to Go 1.19

jbristowe (approved):

PR #53, opened 2022-08-22
Feat/create account links and automation cmd generation

domenicsim1:


pkg/cmd/account/aws/create/create.go
```diff
@@ -50,39 +67,41 @@ func NewCmdCreate(f factory.Factory) *cobra.Command {
...
-		RunE: func(cmd *cobra.Command, args []string) error {
+		RunE: func(cmd *cobra.Command, _ []string) error {
 			client, err := f.GetSpacedClient()
 			if err != nil {
 				return err
 			}
+			opts.CmdPath = cmd.CommandPath()
 			opts.Octopus = client
+			opts.Host = f.GetCurrentHost()
+			opts.Space = f.GetCurrentSpace().GetID()
```
borland, line 78: (Preface: this is my fault, not yours) This is yuck here how you have to call GetCurrentSpace() but it's connected to the fact that you may or may not have called GetSpacedClient earlier. Perhaps we should change GetSpacedClient to return both the client and the space so there's no connascence here.

borland (approved):

PR #54, opened 2022-08-22
Orion/create release 5

borland:
Ticking off some remaining items for create release

### DIFF
--- a/pkg/cmd/release/create/create.go
+++ b/pkg/cmd/release/create/create.go
@@ -26,17 +26,17 @@ import (
 )
 
 const (
-	FlagProject                = "project"
-	FlagPackageVersion         = "package-version" // would default-package-version? be a better name?
-	FlagReleaseNotes           = "release-notes"   // should we also add release-notes-file?
-	FlagChannel                = "channel"
-	FlagVersion                = "version"
-	FlagGitRef                 = "git-ref"
-	FlagGitCommit              = "git-commit"
-	FlagIgnoreExisting         = "ignore-existing"
-	FlagIgnoreChannelRules     = "ignore-channel-rules"
-	FlagPackagePrerelease      = "prerelease-packages"
-	FlagPackageVersionOverride = "package"
+	FlagProject            = "project"
+	FlagPackageVersion     = "package-version"
+	FlagReleaseNotes       = "release-notes" // should we also add release-notes-file?
+	FlagChannel            = "channel"
+	FlagVersion            = "version"
+	FlagGitRef             = "git-ref"
+	FlagGitCommit          = "git-commit"
+	FlagIgnoreExisting     = "ignore-existing"
+	FlagIgnoreChannelRules = "ignore-channel-rules"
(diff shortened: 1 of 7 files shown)


pkg/cmd/release/create/create.go
```diff
@@ -20,54 +21,87 @@ import (
...
 	// stringSlice also allows comma-separated things
-	cmd.Flags().StringSliceP(FlagPackageVersionOverride, "", []string{}, "Version Override for a specific package.\nFormat as {step}:{package}:{version}\nYou may specify this multiple times")
+	flags.StringSliceP(FlagPackageVersionSpec, "", []string{}, "Version specification a specific packages.\nFormat as {package}:{version}, {step}:{version} or {package-ref-name}:{packageOrStep}:{version}\nYou may specify this multiple times")
 
 	// we want the help text to display in the above order, rather than alphabetical
-	cmd.Flags().SortFlags = false
+	flags.SortFlags = false
+
+	// flags aliases for compat with old .NET CLI
+	util.AddFlagAliasesString(flags, FlagGitRef, FlagAliasGitRefRef, FlagAliasGitRefLegacy)
```
borland, line 99: @domenicsim1 this way of doing flag aliases is terrible; if you can figure out a more elegant way I'm keen for it. I tried and failed.

StackOverflow (and the cobra docs themselves somewhere) say that you can do this nicely with `setNormalizeFunc`, however I burned hours on that and it never worked :-(

https://stackoverflow.com/questions/64557741/how-to-support-an-old-flag-name-when-changing-to-a-new-flag-name-with-cobra

Viper apparently proper support, so we might get that if we pull in viper for storing config files?
https://github.com/spf13/viper#registering-and-using-aliases

pkg/cmd/release/create/create.go
```diff
@@ -864,6 +853,97 @@ func AskQuestions(octopus *octopusApiClient.Client, stdout io.Writer, asker ques
...
+			}
+
+			return nil // good!
+		}))
+
+		// if validators return an error, survey retries itself; the errors don't end up at this level.
+		if err != nil {
+			return nil, nil, err
+		}
+		if answer == "u" { // undo!
```
borland, line 924: sneaking in a bonus feature cos it's easy

test/testutil/fakesurvey.go
```diff
@@ -54,62 +64,93 @@ func (m *AskMocker) AsAsker() func(p survey.Prompt, response interface{}, opts .
...
 type QuestionWrapper struct {
 	// in case you need it
 	Question survey.Prompt
+	Options  *survey.AskOptions
 	Asker    *AskMocker
 }
 
-func (q *QuestionWrapper) AnswerWith(answer any) {
-	q.Asker.SendAnswer(answer, nil)
+// AnswerWith runs any validators associated with the question. If they all pass, it sends the answer
```
borland, line 125: This is significant; we now have unit test coverage which covers survey response validators too.
it may not work quite how you may expect, but it's definitely easier and effective this way

domenicsim1 (approved):

PR #57, opened 2022-08-24
feat: custom help page

domenicsim1:


jbristowe (approved):

pkg/cmd/root/help.go
```diff
@@ -0,0 +1,154 @@
...
+			infrastructureCmds = append(infrastructureCmds, s)
+		} else {
+			additionalCmds = append(additionalCmds, s)
+		}
+	}
+
+	// Assume all additional cmds are core if no core cmds are found
+	if len(coreCmds) == 0 {
+		coreCmds = additionalCmds
+		additionalCmds = []string{}
```
borland, line 49: goland always complains at me for empty slice literals and wants me to use `make([]string, 0)` instead. I haven't looked into why.

I don't know if that's useful feedback, but 🤷

pkg/cmd/root/help.go
```diff
@@ -0,0 +1,154 @@
...
+	if isRootCmd(cmd) {
+		helpEntries = append(helpEntries, helpEntry{"LEARN MORE", `
+  Use "octopus [command] [subcommand] --help" for information about a command.`})
+		helpEntries = append(helpEntries, helpEntry{"FEEDBACK", "Open an issue: https://github.com/OctopusDeploy/cli/issues"})
+	}
+
+	out := cmd.OutOrStdout()
+	for _, e := range helpEntries {
+		if e.Title != "" {
+			fmt.Fprintln(out, output.Bold(e.Title))
```
borland, line 107: Super minor: rather than grabbing `cmd.OutOrStdout`, I've been calling `cmd.Printf` and friends (which are just wrappers that cobra has for you). Not sure if that's better or not 🤷

pkg/cmd/root/help.go
```diff
@@ -0,0 +1,154 @@
...
+		fmt.Fprintln(out)
+	}
+}
+
+func rpad(s string, padding int) string {
+	template := fmt.Sprintf("%%-%ds ", padding)
+	return fmt.Sprintf(template, s)
+}
+
+func dedent(s string) string {
```
borland, line 121: Is this removing leading whitespace from a multiline string? I'm not entirely clear on what's going on from the low level character munging; a function-comment would help

borland (approved):

PR #59, opened 2022-08-25
Orion/create release 8

borland:
feat: release create gains outputformat json and basic
feat: If a release version is tied to a donor package, lock that as the release version and just let them add +metadata
feat: fancy help screen for package question loop
feat: can reset package loop
fix: uses git ref canonical name when addressing version controlled projects

### DIFF
--- a/pkg/cmd/release/create/create.go
+++ b/pkg/cmd/release/create/create.go
@@ -312,7 +312,6 @@ func createRun(cmd *cobra.Command, f factory.Factory, flags *CreateFlags) error 
 					cmd.Printf("Successfully created release version %s\n", releaseVersion)
 				}
 			}
-
 		}
 
 		// the API response doesn't tell us what channel it selected, so we need to go look that up to tell the end user
@@ -960,7 +959,6 @@ func AskQuestions(octopus *octopusApiClient.Client, stdout io.Writer, asker ques
 			return cliErrors.NewInvalidResponseError(fmt.Sprintf("cannot determine versioning strategy for project %s", selectedProject.Name))
 		}
 
-		defaultNextVersion := ""
 		if versioningStrategy.DonorPackageStepID != nil || versioningStrategy.DonorPackage != nil {
 			// we've already done the package version work so we can just ask the donor package which version it has selected
 			var donorPackage *StepPackageVersion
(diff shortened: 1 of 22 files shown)


pkg/cmd/release/create/create.go
```diff
@@ -69,6 +72,47 @@ const (
...
+var packageOverrideLoopHelpText = heredoc.Doc(`
+bold(PACKAGE SELECTION)
+ This screen presents the list of packages used by your project, and the steps
+ which reference them. 
+ If an item is dimmed (gray text) this indicates that the attribute is duplicated.
+ For example if you reference the same package in two steps, the second will be dimmed. 
+
+bold(COMMANDS)
+ Any any point, you can enter one of the following:
+ - green(?) to access this help screen
```
borland, line 84: This is like the world's shittiest version of markdown, but the "bang-for-buck" we get with a handful of regexes is great.
FYI @domenicsim1 @jbristowe if you want to re-use the prettifyHelp function in other places

pkg/cmd/release/create/create.go
```diff
@@ -908,31 +975,31 @@ func AskQuestions(octopus *octopusApiClient.Client, stdout io.Writer, asker ques
 				return fmt.Errorf("internal error: can't find donor package in deployment process template - version controlled configuration file in an invalid state")
 			}
 
-			defaultNextVersion = donorPackage.Version
+			versionMetadata, err := askVersionMetadata(asker, donorPackage.PackageID, donorPackage.Version)
```
borland, line 978: @slewis74 merry christmas

pkg/cmd/release/create/create.go
```diff
@@ -1066,15 +1138,40 @@ outer_loop:
...
+		return "", err
+	}
+	return result, nil
+}
+
+func askVersionMetadata(ask question.Asker, packageId string, packageVersion string) (string, error) {
+	var result string
+	if err := ask(&survey.Input{
+		Default: "",
+		Message: fmt.Sprintf("Using release version %s from package %s. Add +metadata? (blank for none)", packageVersion, packageId),
```
borland, line 1155: TODO: I struggled to come up with a sensible (but not too long) prompt here.
The root problem is that I expect statistically zero people to understand what version metadata is
We should discuss this with the team.

pkg/cmd/release/create/create_test.go
```diff
@@ -1153,7 +1307,71 @@ func TestReleaseCreate_AutomationMode(t *testing.T) {
 			_, err = testutil.ReceivePair(cmdReceiver)
 			assert.Nil(t, err)
 
-			assert.Equal(t, "Successfully created release version 1.2.3 (Releases-999) using channel Alpha channel (Channels-32)\n", stdOut.String())
+			assert.Equal(t, heredoc.Doc(`
+				Successfully created release version 1.2.3 using channel Alpha channel
```
borland, line 1311: Note I took out the ID's so there's no more Releases-999 and Channels-32

pkg/cmd/release/list/list.go
```diff
@@ -135,6 +135,7 @@ func listRun(cmd *cobra.Command, f factory.Factory, flags *ListFlags) error {
 
 	return output.PrintArray(allReleases, cmd, output.Mappers[ReleaseViewModel]{
 		Json: func(item ReleaseViewModel) any {
+			// TODO should the ReleaseID go in the JSON as well as the release number? How about channelID?
```
borland, line 138: I'd like to discuss this.

We've decided to steer away from ID's, and now the "human readable" output just says the channel name and release version string. -f basic just says the version string.

But I feel like the JSON should still contain the ID. 🔥🔥🔥??
At the moment it is halfway housed, we get the channel ID, but not the release ID

(Note this applies to both `release list` and `release create`, they output the same structure)

pkg/cmd/release/create/create.go
```diff
@@ -1042,20 +1109,25 @@ outer_loop:
...
-		}
-
-		if resolvedOverride != nil {
+		} else if answer == "r" { // reset! All the way back to the calculated versions, discarding even the stuff that came in from the cmdline
+			if len(packageVersionOverrides) > 0 {
+				packageVersionOverrides = make([]*PackageVersionOverride, 0)
+				overriddenPackageVersions = ApplyPackageOverrides(packageVersionBaseline, packageVersionOverrides)
+			}
+			continue // print table and go again
+		} else if resolvedOverride != nil {
```
domenicsim1, line 1130: This is starting to get bit big, maybe time to use a switch instead. They don’t have problems like falling through in go.

pkg/cmd/release/create/create.go
domenicsim1: Don’t mind the length too much. Got any ideas in how we could make it shorter @slewis74 ?

domenicsim1 (approved):

pkg/cmd/release/create/create.go
borland: I did that internally, but that it changes the semantics of `break` within the 'yes' section... so I'd need an outer break with a label... and then it's not clear that's an improvement anymore

borland: actually turns out we already had an outer break label, so the penalty doesn't apply; I've changed to a switch now and it's better, thanks

PR #64, opened 2022-08-31
Orion/deploy release 1

borland:


go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/OctopusDeploy/cli
 
 go 1.19
 
+replace github.com/OctopusDeploy/go-octopusdeploy/v2 => ../go-octopusdeploy
```
borland, line 5: NOTE: I promise not to merge this.
The reason I've left it in is because the CLI depends on the yet-to-be-merged PR from go-octopusdeploy.

After that merges, we'll need to create a new release (2.x.x) then update this go mod file to point to that instead, THEN this PR can merge

pkg/cmd/release/create/create.go
```diff
@@ -169,7 +170,7 @@ func NewCmdCreate(f factory.Factory) *cobra.Command {
 	flags.StringVarP(&createFlags.PackageVersion.Value, createFlags.PackageVersion.Name, "", "", "Default version to use for all Packages")
 	flags.StringVarP(&createFlags.ReleaseNotes.Value, createFlags.ReleaseNotes.Name, "n", "", "Release notes to attach")
 	flags.StringVarP(&createFlags.ReleaseNotesFile.Value, createFlags.ReleaseNotesFile.Name, "", "", "Release notes to attach (from file)")
-	flags.StringVarP(&createFlags.Version.Value, createFlags.Version.Name, "v", "", "Override the Release Version")
```
borland, line 172: As discussed, removed 'v' short flag from release create version

pkg/cmd/release/create/create_test.go
```diff
@@ -1212,6 +1212,7 @@ func TestReleaseCreate_AutomationMode(t *testing.T) {
 			})
 
 			api.ExpectRequest(t, "GET", "/api").RespondWith(rootResource)
+			api.ExpectRequest(t, "GET", "/api/Spaces-1").RespondWith(rootResource)
```
borland, line 1215: I had to pass a spaceID into the mock, which means that the go-octopusdeploy client kicks in and tries to load a spaced root resource... so the unit tests need to satisfy that. sigh (but it's not a big deal, as you can see here)

pkg/question/selectors/environments.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/environments"
 )
 
-func EnvironmentsMultiSelect(ask question.Asker, client *client.Client, s factory.Spinner, message string) ([]string, error) {
```
borland, line 10: @domenicsim1 
I changed this to return objects rather than ID's because we want a list of environment _names_ in DeployRelease rather than ID's, but I didn't want to break your existing environment code that was using ID's

PR #65, opened 2022-09-02
Package distribution

hnrkndrssn:
This PR implements the package distribution for the CLI.

This includes:
- APT/RPM
- Chocolatey
- Homebrew


.github/workflows/go-releaser.yml
```diff
@@ -12,28 +12,248 @@ permissions:
...
+        printf "::set-output name=zip::%s\n" *.zip
+        unzip -o *.zip && rm -v *.zip
+
+    - name: Setup MSBuild
+      id: setupmsbuild
+      uses: microsoft/setup-msbuild@v1.0.3
+
+    - name: Build MSI
+      id: buildmsi
+      shell: bash
```
borland, line 95: bash on windows?

.github/workflows/go-releaser.yml
```diff
@@ -12,28 +12,248 @@ permissions:
...
+              -kvi "${{ secrets.AZURE_KEYVAULT_CLIENT_ID }}" `
+              -kvs "${{ secrets.AZURE_KEYVAULT_CLIENT_SECRET }}" `
+              -kvc ${{ secrets.AZURE_KEYVAULT_CERTIFICATE_NAME }} `
+              -d "Octopus CLI" `
+              -du "https://octopus.com" `
+              -tr $url `
+              -v `
+              $env:MSI_FILE
+
+            $signSuccessful = $true
```
borland, line 142: can we just 'break' after $signSuccessful = $true?

.github/workflows/go-releaser.yml
```diff
@@ -12,28 +12,248 @@ permissions:
...
+            "http://www.startssl.com/timestamp",
+            "http://timestamp.verisign.com/scripts/timstamp.dll",
+            "http://timestamp.globalsign.com/scripts/timestamp.dll",
+            "https://rfc3161timestamp.globalsign.com/advanced"
+        )
+
+        $ex = $null
+        $signSuccessful = $false
+
+        foreach ($url in $timestampurl) {
```
borland, line 127: This is going to try and sign the MSI using the first timestamp server in the list (comodo), then if that fails for some reason it'll try globalsign, and so forth.

This means that dependent on the transient availability of timestamp servers, different versions of our MSI might be timestamped by different authorities.

I guess it probably doesn't matter given that it doesn't matter who does the timestamping so long as windows trusts that authority? Seems a touch weird is all

.github/workflows/go-releaser.yml
```diff
@@ -12,28 +12,248 @@ permissions:
...
+
+    - name: Download nuget.exe
+      run: curl -o ./nuget.exe https://dist.nuget.org/win-x86-commandline/latest/nuget.exe
+
+    - name: Create Chocolatey package
+      env:
+        MSI_FILE: ${{ needs.msi.outputs.msi_file }}
+      run: |
+        cp artifacts/$(basename "${MSI_FILE//\\//}") build/tools
+        mono nuget.exe pack ./build/cli.nuspec -Version ${{ needs.goreleaser.outputs.version }} -OutputDirectory artifacts/
```
borland, line 198: mono? should this be dotnet?

.github/workflows/go-releaser.yml
```diff
@@ -12,28 +12,248 @@ permissions:
...
+
+    - name: Compress packages
+      run: |
+
+        zip -rj artifacts/octopus-cli.${{ needs.goreleaser.outputs.version }}.zip \
```
borland, line 203: given that this appears to be uploaded to the github release, perhaps we should give it a more descriptive name like `octopus-cli-allplatforms-bundled.zip` or something so that random passersby don't attempt to download it?

.github/workflows/go-releaser.yml
```diff
@@ -12,28 +12,248 @@ permissions:
...
+    - name: Create a release in Octopus Deploy 🐙
+      uses: OctopusDeploy/create-release-action@v2
+      if: "!contains(needs.goreleaser.outputs.version, '-')"
+      with:
+        space: Integrations
+        project: 'cli'
+        package_version: ${{ needs.goreleaser.outputs.version }}
+        packages: '*:NuGet.CommandLine:4.4.1'
+        release_notes_file: ${{ steps.fetch-release-notes.outputs.release-note-file || ''}}
+        git_ref: ${{ (github.ref_type == 'tag' && github.event.repository.default_branch ) || (github.head_ref || github.ref) }}
```
borland, line 258: Am I reading this right? It looks like

git ref = 
   if github.ref_type is 'tag' && the repo has a default branch 
      then use the repo default branch
   else if github.head_ref is set 
     then use github.head_ref
   else 
      use github.ref
      
Why don't we just use github.ref all the time?

.github/workflows/go-releaser.yml
hnrkndrssn: This is not uploaded to the GitHub release, only as an artifact on the build and then pushed to Octopus.

.github/workflows/go-releaser.yml
hnrkndrssn: `nuget.exe` is a full-fat .NET application and does not support `dotnet` only `mono`

.github/workflows/go-releaser.yml
hnrkndrssn: I doubt this would cause any issues, and it's also what we do for the current C# CLI

.github/workflows/go-releaser.yml
hnrkndrssn: @slewis74 do you remember the reason why we had to do this logic, I can't remember the reason but know we use it in most (all) our other builds as well...

.octopus/deployment_process.ocl
```diff
@@ -0,0 +1,192 @@
...
+                
+                username="$(get_octopusvariable 'Publish:HomeBrew:Username')"
+                email="$(get_octopusvariable 'Publish:HomeBrew:UserEmail')"
+                personalAccessToken="$(get_octopusvariable 'Publish:HomeBrew:ApiKey')"
+
+                orgName="OctopusDeploy"
+                repoName="$(get_octopusvariable 'Publish:HomeBrew:RepoName')"
+                releaseAssets=$(curl -X GET https://api.github.com/repos/$orgName/$repoName/releases | jq "map(select(.name == \"$packageVersion\"))[0].assets")
+                
+                # https://github.com/Homebrew/install/#install-homebrew-on-macos-or-linux
```
borland, line 75: sunk cost means it's too late to do this now, but in future because Homebrew depends on ruby, we're guaranteed to have ruby available to us so we could use ruby to do all the work, rather than writing a pile of hard to read `curl | jq` :shrug:

.octopus/deployment_process.ocl
```diff
@@ -0,0 +1,192 @@
...
+                cd cli || exit
+
+                # Deploy APT repo
+                docker run --rm --volume $(pwd):/working \
+                  --env PUBLISH_LINUX_EXTERNAL="$(get_octopusvariable "Publish:Linux:External")" \
+                  --env PUBLISH_ARTIFACTORY_USERNAME="$(get_octopusvariable "Publish:Artifactory:Username")" \
+                  --env PUBLISH_ARTIFACTORY_PASSWORD="$(get_octopusvariable "Publish:Artifactory:Password")" \
+                  --env AWS_ACCESS_KEY_ID="$(get_octopusvariable "LinuxPackagePublisherAwsAccount.AccessKey")" \
+                  --env AWS_SECRET_ACCESS_KEY="$(get_octopusvariable "LinuxPackagePublisherAwsAccount.SecretKey")" \
+                  octopusdeploy/publish-linux bash -c 'cd /working && bash publish-apt.sh' 2>&1 || exit
```
borland, line 136: It looks like this docker image is just ubuntu.. but the runner is already ubuntu... what is docker providing for us here?

.octopus/deployment_process.ocl
```diff
@@ -0,0 +1,192 @@
...
+                cd cli || exit
+
+                # Deploy RPM repo
+                docker run --rm --volume $(pwd):/working \
+                  --env PUBLISH_LINUX_EXTERNAL="$(get_octopusvariable "Publish:Linux:External")" \
+                  --env PUBLISH_ARTIFACTORY_USERNAME="$(get_octopusvariable "Publish:Artifactory:Username")" \
+                  --env PUBLISH_ARTIFACTORY_PASSWORD="$(get_octopusvariable "Publish:Artifactory:Password")" \
+                  --env AWS_ACCESS_KEY_ID="$(get_octopusvariable "LinuxPackagePublisherAwsAccount.AccessKey")" \
+                  --env AWS_SECRET_ACCESS_KEY="$(get_octopusvariable "LinuxPackagePublisherAwsAccount.SecretKey")" \
+                  octopusdeploy/publish-linux bash -c 'cd /working && bash publish-rpm.sh' 2>&1 || exit
```
borland, line 174: as above, what is docker doing here?

.octopus/deployment_settings.ocl
```diff
@@ -0,0 +1,16 @@
+deployment_changes_template = <<-EOT
+        #{each release in Octopus.Deployment.Changes}
+        #{release.ReleaseNotes}
+        #{/each}
+    EOT
```
borland, line 5: misaligned whitespace?

build/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,13 @@
+$ErrorActionPreference = 'Stop'
+
+$toolsDir = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
+$logMsi = Join-Path -Path $env:TEMP -ChildPath ("{0}-{1}-MsiInstall.log" -f $env:ChocolateyPackageName, $env:chocolateyPackageVersion)
+
+$packageArgs = @{
+    packageName    = $env:ChocolateyPackageName
+    fileType       = 'MSI'
+    silentArgs     = "/qn /norestart /l*v `"$logMsi`""
```
borland, line 9: l*v writes the verbose windows installer log to the file `$logMsi`

The verbose windows installer log can be quite big, and is never typically needed unless you're trying to troubleshoot something. 

Should we be enabling this?

build/windows/octopus.wxs
```diff
@@ -0,0 +1,77 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<?ifndef ProductVersion?>
+<?error ProductVersion property not defined?>
+<?endif?>
+
+<!-- Define a unique UpgradeCode per platform -->
+<?if $(var.Platform) = "x64"?>
+<?define InstallerVersion = "200"?>
+<?define UpgradeCode = "738e840f-6305-4a12-8136-8aad988daf43"?>
```
borland, line 10: Just sanity checking; I assume you replaced these with random GUID's? They don't appear to match the github CLI wxs file anyway

build/windows/octopus.wxs
```diff
@@ -0,0 +1,77 @@
...
+<?ifndef ProductVersion?>
+<?error ProductVersion property not defined?>
+<?endif?>
+
+<!-- Define a unique UpgradeCode per platform -->
+<?if $(var.Platform) = "x64"?>
+<?define InstallerVersion = "200"?>
+<?define UpgradeCode = "738e840f-6305-4a12-8136-8aad988daf43"?>
+<?define ProgramFilesFolder = "ProgramFiles64Folder"?>
+<?elseif $(var.Platform) = "x86"?>
```
borland, line 12: We don't have a windows 32 bit intel go binary, so should take this out.

build/windows/octopus.wxs
```diff
@@ -0,0 +1,77 @@
...
+<?define InstallerVersion = "200"?>
+<?define UpgradeCode = "7154b20d-4082-4d7a-9709-00c73dd57072"?>
+<?define ProgramFilesFolder = "ProgramFilesFolder"?>
+<?elseif $(var.Platform) = "arm64"?>
+<?define InstallerVersion = "500"?>
+<?define UpgradeCode = "48c23b4e-ce2b-445b-b4b2-40420f6b10f9"?>
+<?define ProgramFilesFolder = "ProgramFiles64Folder"?>
+<?elseif $(var.Platform) = "arm"?>
```
borland, line 20: We don't have a windows 32 bit arm go binary, so should take this out.

build/windows/octopus.wxs
```diff
@@ -0,0 +1,77 @@
...
+
+<Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
+    <Product Id="*" Name="Octopus CLI" Version="$(var.ProductVersion)" Language="1033" Manufacturer="Octopus Deploy Pty. Ltd." UpgradeCode="$(var.UpgradeCode)">
+        <Package Compressed="yes" InstallerVersion="$(var.InstallerVersion)" InstallScope="perMachine"/>
+        <MediaTemplate EmbedCab="yes"/>
+
+        <!-- Remove older product(s) early but within the transaction -->
+        <MajorUpgrade Schedule="afterInstallInitialize" DowngradeErrorMessage="A newer version of !(bind.property.ProductName) is already installed."/>
+
+        <!-- Upgrade older x86 products -->
```
borland, line 34: Leftover from github cli; we don't have 32 bit things so should take this out

build/windows/octopus.wxs
```diff
@@ -0,0 +1,77 @@
...
+                <File Name="octopus.exe"/>
+                <Environment Id="Path" Action="set" Name="PATH" Part="last" System="yes" Value="[INSTALLDIR]"/>
+            </Component>
+
+            <!-- Persist the INSTALLDIR and restore it in subsequent installs -->
+            <Component Directory="INSTALLDIR">
+                <RegistryValue Root="HKLM" Key="SOFTWARE\Octopus\CLI" Name="InstallDir" Type="string" Value="[INSTALLDIR]"/>
+            </Component>
+
+            <Component Id="OlderX86Env" Guid="32d8ce4c-6422-4f40-8382-cb3ba306b3d1" Directory="TARGETDIR" Win64="no">
```
borland, line 62: as above, should remove

borland (changes requested):

.github/workflows/go-releaser.yml
borland: ok, all good. Will mark conv as resolved

.octopus/deployment_process.ocl
hnrkndrssn: It comes with a couple of extra tools pre-installed that's used by the `publish-*.sh` scripts https://github.com/OctopusDeploy/publish-linux-docker/blob/main/Dockerfile

.github/workflows/go-releaser.yml
borland: Did we consider using `dotnet pack` instead of nuget.exe?
Or just plain old `zip`?

I guess it doesn't matter anyway if mono is already available on the linux VM and it doesn't cost us to install it 🤷

.octopus/deployment_settings.ocl
hnrkndrssn: That's straight out of Octopus after CaC'ifying a project

.github/workflows/go-releaser.yml
borland: ah, right, thanks!

.github/workflows/go-releaser.yml
hnrkndrssn: chocolatey does not support `zip` files, `dotnet pack` does not produce the correct format package from memory

build/windows/octopus.wxs
hnrkndrssn: Yes, we replaced these GUIDs

.github/workflows/go-releaser.yml
borland: I was referencing the fact that a nupkg file is just a zip with a predetermined file structure; you can make one manually with plain old zip. But yeah, if we aren't paying a cost to install mono (it's quite large) then doesn't matter

.octopus/deployment_settings.ocl
borland: LOL :-)

.github/workflows/go-releaser.yml
hnrkndrssn: > a nupkg file is just a zip with a predetermined file structure

Yes, but `nuget pack` also includes a bunch of other files that we'd have to manually construct and include.

.octopus/deployment_process.ocl
hnrkndrssn: I agree that all the `jq` calls makes it yuck, so I simplified it to just string "interpolation" as the URL is well-known

build/tools/chocolateyInstall.ps1
hnrkndrssn: I've removed the verbose logging, we can re-enable at a later time if needed.

.github/workflows/go-releaser.yml
hnrkndrssn: So this is why:
- For tags, `github.ref` is the tag ref i.e. `refs/tags/<tag_name>` so we'd have to use the default branch.
- For pull requests, `github.ref` is the merge branch for the pull request i.e. `refs/pull/<pr_number>/merge` so we have to use `github.head_ref` as it's the branch of the PR

In this workflow, we can simplify this by just using `github.default_branch` as the workflow is only fired when a tag is pushed.

mieliespoor:
Any change of adding winget support?

hnrkndrssn:
> Any chance of adding winget support?

It's on the cards for the future, just not yet sorry! 😊

borland (approved):

PR #66, opened 2022-09-07
feat: config file support

### LINKED ISSUE #62: Configuration
Create a file to easily persist common configuration.
This will be achieved using [Viper](https://github.com/spf13/viper).

The following keys are being considered to include in the configuration file.
| Key | Default Value | Description |
| ---  | ------------  | ------------ |
|  Host | | Octopus Deploy Host |
| ApiKey | | Octopus Deploy API key. Note stored as plain text. |
| Space | | An Octopus  space to use |
| NoPrompt | disabled | Disable interactive mode if true |
| ProxyUrl | | Set proxy to direct traffic through |
| Editor | unix: nano, windows: notepad | The text editor program to use for authoring text |
| ShowOctopus | enabled for first run only | Show octopus on root help command |
| OutputFormat | table | Output format to enforce |

Configuration file will be stored in the following locations:

- Window: %APPDATA%\Octopus\cli_config.<filetype>
- Unix: ~/.config/octopus/cli_config.<filetype>

Configuration will follow the precedence order enforce by [Viper](https://github.com/spf13/viper#why-viper). Each item takes precedence over the item below it:

1. flag
2. env
3. config
4. key/value store (config file)
5. default 

The following new commands will be created to easily interact with the configuration file:

- octopus config set <key> <value>
  - set value for key in config file
- octopus config list [flags]
  - list all values on config file 
  - supports `output-format` flag
- octopus config get <key>
  - show value for given key

### Interactive Mode

`octopus config set`
```
? Select the key you would you like to change?
> Host
  Api Key
  Space
...
```

`octopus config set host`
```
? Enter new value for Host: <input>
```

`octopus config get`
```
? Select the key you would you like to see the value of?
> Host
  Api Key
  Space
...
```

domenicsim1:
Closes #62


pkg/config/config.go
```diff
@@ -0,0 +1,119 @@
...
+	//	viper.SetDefault(constants.ConfigShowOctopus, true)
+	viper.SetDefault(constants.ConfigOutputFormat, "table")
+
+	viper.SetConfigName(configName)
+	viper.SetConfigType(defaultConfigFileType)
+
+	if runtime.GOOS == "windows" {
+		viper.SetDefault(constants.ConfigEditor, "notepad")
+	} else { // unix
+		viper.SetDefault(constants.ConfigEditor, "nano")
```
borland, line 35: Nice how this interacts with the bind env stuff later and survey editor!

pkg/config/config.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	getConfigPath()
```
borland, line 37: useless call to `getConfigPath()`

pkg/config/config.go
```diff
@@ -0,0 +1,119 @@
...
+
+	if err := viper.ReadInConfig(); err != nil {
+		if _, ok := err.(viper.ConfigFileNotFoundError); ok {
+			// Do nothing, config file will be created on `config set` cmd
+			// This is to avoid issues with CI tools that may not have access
+			// to the file system
+		} else {
+			// Config file was found but something is wrong
+			panic(fmt.Errorf("fatal error reading config file: %w", err))
```
borland, line 46: Rather than panic, IMHO it'd be nicer to print the error but then allow the CLI to proceed with a default config setup

pkg/config/config.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+	}
+}
+
+func SetupEnv() {
+	viper.BindEnv(constants.ConfigApiKey, constants.EnvOctopusApiKey)
+	viper.BindEnv(constants.ConfigHost, constants.EnvOctopusHost)
+	viper.BindEnv(constants.ConfigSpace, constants.EnvOctopusSpace)
+	viper.BindEnv(constants.ConfigEditor, constants.EnvEditor, constants.EnvVisual)
```
borland, line 55: q: if both EDITOR and VISUAL are set, which one wins here? Might be useful to add a comment in the code explaining this

pkg/config/config.go
```diff
@@ -0,0 +1,119 @@
...
+
+func writeNewConfig() error {
+	configPath, err := getConfigPath()
+	if err != nil {
+		return err
+	}
+	if err := os.MkdirAll(configPath, os.ModePerm); err != nil {
+		return err
+	}
+	return viper.SafeWriteConfigAs(fmt.Sprintf("%s/%s.%s", configPath, configName, defaultConfigFileType))
```
borland, line 107: Use filepath.Join rather than hardcoding `/` as a directory separator

pkg/config/config.go
```diff
@@ -0,0 +1,119 @@
...
+		} else {
+			return "", fmt.Errorf("error could not find path to appdata")
+		}
+	}
+	// is unix
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return "", fmt.Errorf("error could not find user home directory: %w", err)
+	}
+	configPath := filepath.Join(home, ".config/octopus")
```
borland, line 94: Use `filepath.Join(home, ".config", "octopus")` rather than hardcoding `/` as a directory separator

pkg/config/config.go
```diff
@@ -0,0 +1,119 @@
...
+	if err := os.MkdirAll(configPath, os.ModePerm); err != nil {
+		return err
+	}
+	return viper.SafeWriteConfigAs(fmt.Sprintf("%s/%s.%s", configPath, configName, defaultConfigFileType))
+}
+
+func ValidateKey(key string) bool {
+	key = strings.TrimSpace(key)
+	key = strings.ToLower(key)
+	for _, v := range viper.AllKeys() {
```
borland, line 113: you can probably use `util.SliceContains` here to avoid the for loop

pkg/cmd/config/config.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	getCmd "github.com/OctopusDeploy/cli/pkg/cmd/config/get"
+	setCmd "github.com/OctopusDeploy/cli/pkg/cmd/config/set"
+	"github.com/OctopusDeploy/cli/pkg/constants/annotations"
+	"github.com/OctopusDeploy/cli/pkg/factory"
+	"github.com/spf13/cobra"
+)
+
+func NewCmdRelease(f factory.Factory) *cobra.Command {
```
borland, line 11: should be `NewCmdConfig`

pkg/cmd/config/config.go
```diff
@@ -0,0 +1,24 @@
...
+		Short: "Manage CLI config file",
+		Long:  `Manage the CLI config file`,
+		Annotations: map[string]string{
+			annotations.IsConfiguration: "true",
+		},
+	}
+
+	cmd.AddCommand(getCmd.NewCmdGet(f))
+	cmd.AddCommand(setCmd.NewCmdSet(f))
+	return cmd
```
borland, line 23: Best done in a follow-up PR, but we should have a 'config list' command too

pkg/cmd/root/root.go
```diff
@@ -56,14 +58,17 @@ func NewCmdRoot(f factory.Factory, clientFactory apiclient.ClientFactory, askPro
 	cmd.AddCommand(environmentCmd.NewCmdEnvironment(f))
 
 	// configuration commands
+	cmd.AddCommand(configCmd.NewCmdRelease(f))
 	cmd.AddCommand(spaceCmd.NewCmdSpace(f))
 
 	cmd.AddCommand(releaseCmd.NewCmdRelease(f))
 
+	viper.BindPFlag(constants.ConfigNoPrompt, cmdPFlags.Lookup(constants.FlagNoPrompt))
```
borland, line 66: question: Does this take the value from the viper config file, and stick it into the pflag? or does it go the other way?

Once we have a viper flag bound to a pflag, should we be using viper to read the value, or the pflag?

The viper help is predictably useless: https://pkg.go.dev/github.com/spf13/viper#BindPFlag

borland (changes requested):

pkg/config/config.go
domenicsim1: this is used to set the config path in viper. Will make this more clear.

pkg/config/config.go
domenicsim1: They will take precedence in the specified order, will leave a comment.

pkg/cmd/root/root.go
domenicsim1: They way it works is when the p flag is bound viper will use it if the value is set. Viper reads the pflag when you call `viper.get` for the key we bind too. Viper will check if the key we request has a bounded pflag first, after that if not flag value is set it will check env, lastly is checks the config file.

domenicsim1: So yes, if we bind a pflag in viper, we should be using viper to request that value

borland (approved):

PR #70, opened 2022-09-08
Orion/deploy release 2

borland:


go.mod
```diff
@@ -2,10 +2,12 @@ module github.com/OctopusDeploy/cli
 
 go 1.19
 
+//replace github.com/OctopusDeploy/go-octopusdeploy/v2 => ../go-octopusdeploy
+
 require (
 	github.com/AlecAivazis/survey/v2 v2.3.5
 	github.com/MakeNowJust/heredoc/v2 v2.0.1
-	github.com/OctopusDeploy/go-octopusdeploy/v2 v2.3.3
+	github.com/OctopusDeploy/go-octopusdeploy/v2 v2.3.4-0.20220906224003-626a6680def2
```
borland, line 10: I can't fix this until someone approves the other PR please
https://github.com/OctopusDeploy/go-octopusdeploy/pull/138

pkg/cmd/account/gcp/create/create.go
```diff
@@ -196,13 +198,13 @@ func promptMissing(opts *CreateOptions) error {
 	}
 
 	if opts.Environments.Value == nil {
-		environmentIDs, err := selectors.EnvironmentsMultiSelect(opts.Ask, opts.Octopus,
+		envs, err := selectors.EnvironmentsMultiSelect(opts.Ask, opts.Octopus,
```
borland, line 201: attention @domenicsim1 this kind of refactoring is why I was asking for at least one or two basic tests over the create account stuff.

go.mod
```diff
@@ -2,10 +2,12 @@ module github.com/OctopusDeploy/cli
 
 go 1.19
 
+//replace github.com/OctopusDeploy/go-octopusdeploy/v2 => ../go-octopusdeploy
```
domenicsim1, line 5: remove this

pkg/cmd/release/deploy/deploy.go
```diff
@@ -0,0 +1,1194 @@
...
+		defTrueFalse := "False"
+		if b, err := strconv.ParseBool(defaultValue); err == nil && b {
+			defTrueFalse = "True"
+		}
+		var response string
+		err := asker(&survey.Select{
+			Message: message,
+			Default: defTrueFalse,
+			Options: []string{"True", "False"}, // Yes/No would read more nicely, but doesn't fit well with cmdline which expects True/False
+		}, &response, askOpt)
```
domenicsim1, line 1083: why did we chose to use select; over `survery.Confirm` and covert the response bool to a string?

domenicsim1 (approved):
Good stuff 👍

pkg/cmd/release/deploy/deploy.go
borland: No strong reason. This whole thing still needs a 'ux' pass with JB and shannon, we might yet change some of this kind of stuff.

I chose a select with a true/false because it feels "easier", you can just sit on the arrow keys and enter rather than having to type 'y' on the keyboard. Also (perhaps more importantly) because the command line flag requires you to specify "true" or "false", whereas confirm is always y/n, so it's - in a small way - nice to present true/false here as well

PR #73, opened 2022-09-12
Orion/release deploy ux

borland:


pkg/cmd/release/deploy/deploy.go
```diff
@@ -227,9 +221,9 @@ func deployRun(cmd *cobra.Command, f factory.Factory, flags *DeployFlags) error
 	}
 
 	if f.IsPromptEnabled() {
-		now := time.Now()
+		now := time.Now
```
borland, line 224: Previously it captured 'now' at the start of the process, so if you took ages answering questions, it would be wrong by the time you got to the scheduled start date. Fixed by capturing a function pointer which we don't call until the last possible time

pkg/util/pflagaliases.go
```diff
@@ -0,0 +1,64 @@
...
+	for k, v := range aliases {
+		primaryFlag := flags.Lookup(k)
+		for _, aliasName := range v {
+			aliasFlag := flags.Lookup(aliasName)
+			aliasValueString := aliasFlag.Value.String() // flags get stringified here, but it's fast enough and a one-shot so meh
+			if aliasValueString != aliasFlag.DefValue {
+				// we have to call set because .Value holds the pointer to the bound variable;
+				// if we set one Value to another we end up pointing at different storage and it doesn't work
+
+				if aliasFlag.DefValue == "[]" && len(aliasValueString) > 2 && aliasValueString[0] == '[' {
```
borland, line 52: This code is 🤮 and will break instantly, should survey change the way they print flags internally, or anything like that. But it's the best we can do with the API surface that survey gives to us; Unit tests will catch it if something ever does break so 🤷

pkg/util/pflagaliases.go
```diff
@@ -0,0 +1,64 @@
...
+					// this is not great. We rely on the assumption that pflag's internal Set(string) calls readAsCsv in a
+					// predictable way that doesn't change. However, there is nothing in the pflag public API that would
+					// allow us to do a better job, as flag values are only exposed via the `Value` interface which
+					// only allows read/write of values using String.
+					_ = primaryFlag.Value.Set(aliasValueString[1 : len(aliasValueString)-1])
```
borland, line 57: What's going on here is that when you read the `.Value.String` of an array flags, it returns "[thing1,thing2]". When you call `.Value.Set` it only accepts either a single value, or a comma separated list, without the surrouding `[]`, so I have to strip them off.

pkg/util/pflagaliases_test.go
```diff
@@ -0,0 +1,141 @@
...
+		assert.Nil(t, someFlagValue)
+
+		util.ApplyFlagAliases(flags, aliases)
+		assert.Equal(t, []string{"someValue", "secondValue", "thirdValue"}, someFlagValue)
+	})
+
+	t.Run("mixed", func(t *testing.T) {
+		var someFlagValue []string
+		flags, aliases := setup(&someFlagValue)
+		assert.Nil(t, flags.Parse([]string{"--some-flag", "initialValue", "--someFlag", "alias1value", "--sf", "alias2value"}))
```
borland, line 133: It's cool that this actually works, I really didn't think it would. (this all depends on internal implementation quirks inside survey that we can't change)

pkg/util/pflagaliases_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+		// sanity check, the value should have been parsed into the hidden "someFlag" flag, but not copied into the primary "some-flag" yet
+		assert.Equal(t, []string{"initialValue"}, someFlagValue)
+
+		util.ApplyFlagAliases(flags, aliases)
+		assert.Equal(t, []string{"initialValue", "alias1value", "alias2value"}, someFlagValue)
```
borland, line 139: It's cool that this actually works, I really didn't think it would. (this all depends on internal implementation quirks inside survey that we can't change)

domenicsim1 (approved):
👍🏻

PR #80, opened 2022-09-16
fix: cli#71 - config system was leaking env vars into the config file when it shouldn't have 

borland:
If you bindEnv/flags etc into viper, then it WILL write them to the config file, by design.

This was already handled by using a secondary viper inside 'config set'; however the global viper would also write the config file on startup as part of it's code to check that the file existed. This is how the env vars leaked into it.

Fix: 
- The global viper can pick up all the env vars and flags and only read. It must NOT write to or create the file
- The local viper inside 'config set' can read, write and create the file, but it must NOT pickup env vars and flags.

Root cause:
The code which created the file as part of reading it was trying to be helpful. It wasn't easy to pick up, as it was nested inside a few layers... Better to be explicit about when the file is created and avoid magic behaviour.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -2,21 +2,18 @@ package config
 
 import (
 	"fmt"
+	"github.com/OctopusDeploy/cli/pkg/constants"
+	"github.com/OctopusDeploy/cli/pkg/util"
+	"github.com/spf13/viper"
 	"os"
 	"path/filepath"
 	"runtime"
 	"strings"
-
-	"github.com/OctopusDeploy/cli/pkg/constants"
-	"github.com/OctopusDeploy/cli/pkg/util"
-	"github.com/spf13/viper"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/config/config.go
```diff
@@ -2,88 +2,105 @@ package config
...
+	return configPath, nil
+}
+
+// getConfigPath works out the directory where the config file should be saved and returns it.
+// does not modify the global viper
 func getConfigPath() (string, error) {
 	if runtime.GOOS == "windows" {
 		if appdataPath := os.Getenv(appData); appdataPath != "" {
 			configPath := filepath.Join(appdataPath, "octopus")
-			viper.AddConfigPath(configPath)
```
borland, line 86: Burying the side-effect of setting the config path on the global viper instance made this pretty hard to figure out :-(

domenicsim1 (approved):

PR #83, opened 2022-09-20
octopus config set interactive mode broken

### LINKED ISSUE #82: octopus config set is broken in interactive mode
Repro on version 0.2.5 of the CLI

Start with a clean system, delete any existing `cli_config.json` from your home folder

```
octopus config set
? What key would you like to change? (select ApiKey)
```

Expected to prompt for input so you can type/paste the api key

Actual: CLI exits with error `unable to get value for key: ApiKey`

This happens for any config value where an entry doesn't already exist in `cli_config.json`

borland:
Fix #82


domenicsim1 (approved):
👍

PR #86, opened 2022-09-23
Orion/push package

borland:


pkg/cmd/package/upload/upload.go
```diff
@@ -0,0 +1,182 @@
...
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/packages"
+	"github.com/spf13/cobra"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// TODO Outstanding questions for package upload:
+// - What should the response be for outputformat basic and json?
```
jbristowe, line 22: The basic output from `octo` looks something like this:

```
Octopus CLI, version 9.x.y

Detected automation environment: "NoneOrUnknown"
Space name unspecified, process will run in the default space context
Handshaking with Octopus Server: https://foobar.octopus.app
Handshake successful. Octopus version: 2022.x.y; API version: 3.0.0
Authenticated as: Foo Bar <foo@bar.com> 
Pushing package: /Users/foobar/foo.1.1.0.zip...
Requesting signature for delta compression from the server for upload of a package with id 'foo' and version '1.1.0'
No package with the same ID exists on the server
Falling back to pushing the complete package to the server
Push successful
```

Here's the output in JSON:

```
{
  "SuccessfulPackages": [
    "/Users/johnbristowe/GitHub/OctopusCLI/foo.1.1.1.zip"
  ],
  "FailedPackages": []
}
```

pkg/cmd/package/upload/upload.go
```diff
@@ -0,0 +1,182 @@
...
+// - Basic output: I have "Successfully uploaded package x-1.0.0.zip" when the package was created/overwritten, and "Successfully processed package x-1.0.0.zip" when it was ignored. Is this right?
```
jbristowe, line 23: `octo` will inform you if the package is NOT overwritten. We should do the same.

pkg/cmd/package/upload/upload.go
```diff
@@ -0,0 +1,182 @@
...
+// - If we specify multiple packages and one of them fails, does it keep going or abort on first error?
```
jbristowe, line 24: `octo` reports successful and failed packages (for upload). We could do the same.

pkg/cmd/package/upload/upload.go
```diff
@@ -0,0 +1,182 @@
...
+// - the package upload response from the server has a load of stuff in it. Is there any value in doing anything with this?
```
jbristowe, line 25: Not at the moment. But perhaps in the future?

pkg/cmd/package/upload/upload.go
borland: ok cool I'll have a play with `octo` and put in some similar wording

pkg/cmd/package/upload/upload.go
borland: cool, I will keep going

pkg/cmd/package/upload/upload.go
borland: Cool, thanks for that. I'll remove this TODO comment and add another one at the callsite explaining that while `UploadPackageResponse` has a lot of data in it, the CLI deliberately doesn't use any at this time

pkg/cmd/package/upload/upload.go
borland: My intent with all these TODOS was to merge this PR first as a 'first pass' that people could play with, then do a second pass and tidy them all up, plus any other UX/wording feedback we gathered.

However, I didn't actually tell anyone about that. Whoops, so of course you'd reject the PR on grounds of unresolved TODOs :-)

domenicsim1 (approved):

PR #88, opened 2022-09-27
Use the GitHub API instead of the GitHub CLI to create the PR

hnrkndrssn:
While using `gh` is nice, the fact that we have to maintain a package that contains it (even if that's just getting the package from the release on GitHub and uploading it to Octopus) makes it a less optimal solution.

The script can still be run locally but will only create the PR for the update if you pass in a GitHub PAT.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -25,9 +25,6 @@ builds:
 archives:
   - replacements:
       darwin: macOS
-      linux: Linux
-      windows: Windows
-      amd64: x86_64
     format_overrides:
       - goos: windows
         format: zip
--- a/.octopus/deployment_process.ocl
+++ b/.octopus/deployment_process.ocl
@@ -66,27 +66,20 @@ step "create-pull-request-to-update-formula-in-homebrew" {
                 param(
                     [String]$packageVersion, # e.g. "0.2.2"
                     [String]$extractedPath # e.g.  "C:\Users\Orion\Downloads\octopus-cli.0.2.2"
+                    [String]$githubtoken
                 )
                 
                 $origin="https://github.com/OctopusDeploy/homebrew-taps"
-                $gh = "gh" # github cli
                 
                 if ($OctopusParameters) {
-                    write-host "using OctopusParameters"
-                    # we are running inside an octopus tentacle
                     $packageVersion = $OctopusParameters["Octopus.Action.Package[cli].PackageVersion"]
                     $extractedPath = $OctopusParameters["Octopus.Action.Package[cli].ExtractedPath"]
                 
-                    $ghExtractedPath = $OctopusParameters["Octopus.Action.Package[gh_linux_amd64].ExtractedPath"]
-                    write-host "dbg: ghExtractedPath = $ghExtractedPath"
-                    $ghVersion = $OctopusParameters["Octopus.Action.Package[gh_linux_amd64].PackageVersion"]
-                    write-host "dbg: ghVersion = $ghVersion"
-                    $gh = Join-Path -Path $ghExtractedPath "gh_${ghVersion}_linux_amd64" "bin" "gh"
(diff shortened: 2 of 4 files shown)


.goreleaser.yaml
```diff
@@ -25,9 +25,6 @@ builds:
 archives:
   - replacements:
       darwin: macOS
-      linux: Linux
-      windows: Windows
-      amd64: x86_64
```
borland, line 30: Was there a reason you changed then naming scheme here?

The change to call this `x86_64` was deliberate, as that is definitely the dominant term on macOS. 
On windows it seems to be mostly `x64` with a bit of `x86_64` thrown in there. 

It's really only linux that prefers `amd64`, which is a bit of a niche, especially given that we expect most linuxy people to go via the deb or rpm anyway.

borland: Blocking on this

.octopus/deployment_process.ocl
```diff
@@ -119,18 +112,20 @@ step "create-pull-request-to-update-formula-in-homebrew" {
...
-                
                 git push --set-upstream origin $branchName
-                
-                &($gh) pr create --head $branchName --title "octopus-cli $packageVersion" --body "automatically generated"
-                
+
+                if ($githubtoken) {
+                    $baseBranch = $(git symbolic-ref "refs/remotes/origin/HEAD") -replace "refs/remotes/origin/", ""
+
+                    curl -X POST `
```
borland, line 122: Does this work? We're inside powershell, which sets an alias for `curl` to equal `Invoke-WebRequest` which is dumb and annoying as almost all the flags aren't the same as curl. Github actions runners do have the win32 versions of bash, curl, etc installed though, so if it lands on that then it'll be ok.

IMHO it'd be cleaner if we just had powershell do the web request natively, and not take any dependencies on linux stuff, it looks like this is pretty trivial
https://www.thecodebuzz.com/making-powershell-post-request-with-body-json-array/

```
Invoke-WebRequest -Method "POST" -Uri https://api.github.com/repos/OctopusDeploy/homebrew-taps/pulls `
-Body.. etc
```

borland (changes requested):

.goreleaser.yaml
hnrkndrssn: I asked the question a while back if there was a particular reason for this replacement and was told there wasn't. I also looked at the GitHub CLI (as we've been using it as a base line for certain things) and they do not have this replacement.

.octopus/deployment_process.ocl
hnrkndrssn: Honestly forgot we were in PowerShell here, so just straight copied the code from the GitHub API docs. Will update to make sure it works in PowerShell

.goreleaser.yaml
borland: 🤷 ok, I'm good with that.

.octopus/deployment_process.ocl
hnrkndrssn: I've converted it to use `Invoke-RestMethod` instead of `curl`

.octopus/deployment_process.ocl
```diff
@@ -119,18 +112,19 @@ step "create-pull-request-to-update-formula-in-homebrew" {
...
-                
                 git push --set-upstream origin $branchName
-                
-                &($gh) pr create --head $branchName --title "octopus-cli $packageVersion" --body "automatically generated"
-                
+
+                if ($githubtoken) {
+                    $baseBranch = $(git symbolic-ref "refs/remotes/origin/HEAD") -replace "refs/remotes/origin/", ""
+
+                    Invoke-RestMethod -Method POST `
```
borland, line 122: Nice, I didn't know about `Invoke-RestMethod`

borland (approved):

PR #105, opened 2022-10-03
Feat/config list

domenicsim1:


pkg/cmd/config/list/list.go
```diff
@@ -0,0 +1,89 @@
...
+		cmd.Println(string(data))
+	case constants.OutputFormatBasic:
+		for _, key := range configFile.AllKeys() {
+			cmd.Println(configFile.GetString(key))
+		}
+	case constants.OutputFormatTable:
+		t := output.NewTable(cmd.OutOrStdout())
+		t.AddRow(output.Bold("KEY"), output.Bold("VALUE"))
+		for _, key := range configFile.AllKeys() {
+			t.AddRow(key, configFile.GetString(key))
```
borland, line 83: This is going to print the API key directly to the screen.
I feel like we should print ****s instead for this (same deal with JSON), but I'm not sure if I'm over-thinking it. 

Right now, the API key is trivially readable from the JSON file, so printing it to the screen isn't much worse -- but we have discussed things like putting the API key in the windows/macos keystore, and printing those values isn't something you'd want to be doing normally.

Would be good to get @jbristowe 's opinion on this one.

borland:
PR looks good, I added a question about whether or not we should print API keys in config list. Deliberately neither approving nor blocking the PR, the team can resolve this question and go from there

pkg/cmd/config/list/list.go
jbristowe: I'm inclined to agree here. Maybe we should do what they do with credit card numbers; show the last 4 digits and print `*` for the rest.

domenicsim1: As already said. It's still plain text in the json and the `config get apikey` command will still return the text value. I hid it anyway in the list command, as I could see "accidental" exposing the  key here as an issue as where doing it as `config get apikey` is it explicit.

jbristowe (approved):

PR #106, opened 2022-10-04
add project list command

benPearce1:
[sc-18973]

### DIFF
--- a/pkg/cmd/project/list/list.go
+++ b/pkg/cmd/project/list/list.go
@@ -1,0 +1,53 @@
+package list
+
+import (
+	"github.com/MakeNowJust/heredoc/v2"
+	"github.com/OctopusDeploy/cli/pkg/factory"
+	"github.com/OctopusDeploy/cli/pkg/output"
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/projects"
+	"github.com/spf13/cobra"
+)
+
+func NewCmdList(f factory.Factory) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "list",
+		Short: "List projects in Octopus Deploy",
+		Long:  "List projects in Octopus Deploy",
+		Example: heredoc.Doc(`
+			$ octopus project list		
+		`),
+		Aliases: []string{"ls"},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return listRun(cmd, f)
+		},
+	}
+
+	return cmd
+}
+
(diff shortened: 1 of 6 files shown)


pkg/cmd/project/project.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/OctopusDeploy/cli/pkg/constants"
+	"github.com/OctopusDeploy/cli/pkg/constants/annotations"
+	"github.com/OctopusDeploy/cli/pkg/factory"
+	"github.com/spf13/cobra"
+)
+
+func NewCmdProject(f factory.Factory) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "project <command>",
+		Aliases: []string{"proj"},
```
borland, line 16: we've not put an alias on a toplevel command before. 
@jbristowe what do you think about this? I'm very slightly leaning towards "let's not do that", but not concerned either way.

pkg/cmd/project/list/list.go
```diff
@@ -0,0 +1,53 @@
+package list
```
borland, line 1: TODO some tests?

borland (commented):
Nice work Ben! I will refrain from approve/reject as I'm supposed to be on nancy :-)

pkg/cmd/project/project.go
jbristowe: Agreed. I do not wish to have aliases for top-level commands. Please remove the `env` alias as well.

pkg/cmd/project/project.go
```diff
@@ -0,0 +1,31 @@
...
+		Short:   "Manage projects",
+		Long:    `Work with Octopus Deploy projects.`,
+		Example: fmt.Sprintf(heredoc.Doc(`
+			$ %s project list
+			$ %s env ls
```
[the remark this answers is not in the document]
benPearce1: 👍 yes, it was a copy-pasta

pkg/cmd/project/project.go
benPearce1: `env` is now removed, no other top-level commands have an alias defined

pkg/cmd/project/list/list.go
benPearce1: still getting to that

benPearce1: actually not required for simple list commands

domenicsim1 (commented):
😎 👍

domenicsim1 (approved):

PR #108, opened 2022-10-05
feat: add project view command

benPearce1:
[sc-18973]

Depends on https://github.com/OctopusDeploy/go-octopusdeploy/pull/145

Help output:
```
View a project in an instance of Octopus Deploy

USAGE
 octopus project view {<name> | <id> | <slug>} [flags]

FLAGS
 -w, --web   Open in web browser

INHERITED FLAGS
 -h, --help                   Show help for command
     --no-prompt              disable prompting in interactive mode
 -f, --output-format string   Output Format (Valid values are 'json', 'table', 'basic'. Defaults to table)
 -s, --space string           Set Space

EXAMPLE
 $ octopus project view Projects-9000
 $ octopus project view 'Deploy Web App'
 $ octopus project view deploy-web-app
```

View command output using the `slug` as the project identifier
```
❯ ./octopus project view with-cac -s Default
with cac (Projects-321)
Version control branch: main
description (is this branch specific)

View this project in Octopus Deploy: https://b.testoctopus.app/app#/Spaces-1/projects/Projects-321
```

With colour formatting:
image

`--web` also successfully opens the project URL in the default web browser

### DIFF
--- a/pkg/cmd/project/view/view.go
+++ b/pkg/cmd/project/view/view.go
@@ -1,0 +1,106 @@
+package view
+
+import (
+	"fmt"
+	"github.com/MakeNowJust/heredoc/v2"
+	"github.com/OctopusDeploy/cli/pkg/constants"
+	"github.com/OctopusDeploy/cli/pkg/factory"
+	"github.com/OctopusDeploy/cli/pkg/output"
+	"github.com/OctopusDeploy/cli/pkg/usage"
+	"github.com/OctopusDeploy/cli/pkg/util/flag"
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/client"
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/projects"
+	"github.com/pkg/browser"
+	"github.com/spf13/cobra"
+	"io"
+	"os"
+)
+
+const (
+	FlagWeb = "web"
+)
+
+type ViewFlags struct {
+	Web *flag.Flag[bool]
+}
+
+func NewViewFlags() *ViewFlags {
+	return &ViewFlags{
+		Web: flag.New[bool](FlagWeb, false),
+	}
+}
+
+type ViewOptions struct {
+	Client   *client.Client
+	Host     string
+	out      io.Writer
+	idOrName string
+	flags    *ViewFlags
+}
+
+func NewCmdView(f factory.Factory) *cobra.Command {
+	viewFlags := NewViewFlags()
+	cmd := &cobra.Command{
+		Args:  usage.ExactArgs(1),
+		Use:   "view {<name> | <id> | <slug>}",
+		Short: "View a project in an instance of Octopus Deploy",
+		Long:  "View a project in an instance of Octopus Deploy",
+		Example: fmt.Sprintf(heredoc.Doc(`
+			$ %s project view Projects-9000
+			$ %s project view 'Deploy Web App'
+			$ %s project view deploy-web-app
+		`), constants.ExecutableName, constants.ExecutableName, constants.ExecutableName),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			client, err := f.GetSpacedClient()
+			if err != nil {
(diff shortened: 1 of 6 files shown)


pkg/cmd/project/view/view.go
```diff
@@ -0,0 +1,106 @@
...
+import (
+	"fmt"
+	"github.com/MakeNowJust/heredoc/v2"
+	"github.com/OctopusDeploy/cli/pkg/constants"
+	"github.com/OctopusDeploy/cli/pkg/factory"
+	"github.com/OctopusDeploy/cli/pkg/output"
+	"github.com/OctopusDeploy/cli/pkg/usage"
+	"github.com/OctopusDeploy/cli/pkg/util/flag"
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/client"
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/projects"
```
borland, line 12: > imported and not used: "github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/projects"

@benPearce1 what IDE are you using?

This kind of problem hasn't happened on a PR before because JetBrains GoLand will automatically fixup any imports every time you save the file, and I think VSCode does the same if the extensions are working properly. 

I find GoLand to be nice but Dom+JB prefer VSCode. Both seem to work well. If you would like a hand getting GoLand up and running sing out and I'd be happy to help

pkg/cmd/project/view/view.go
```diff
@@ -0,0 +1,106 @@
...
+	return cmd
+}
+
+func viewRun(opts *ViewOptions) error {
+	project, err := opts.Client.Projects.GetByIdOrName(opts.idOrName)
+	if err != nil {
+		return err
+	}
+
+	fmt.Fprintf(opts.out, "%s %s\n", output.Bold(project.Name), output.Dimf("(%s)", project.GetID()))
```
borland, line 83: I think this is going to throw some lint warnings because you're not consuming the return values from `fmt.Fprintf`. Off the top of my head it has two of them

pkg/cmd/space/view/view.go
```diff
@@ -44,7 +44,7 @@ func NewCmdView(f factory.Factory) *cobra.Command {
 			}
 
 			opts.Client = client
-			opts.Host = os.Getenv("OCTOPUS_HOST")
+			opts.Host = os.Getenv(constants.EnvOctopusHost)
```
borland, line 47: I realise this is not your code @benPearce1 and is leftover from some earlier CLI code, no fault of yours.
Nevertheless this is a bug.

The `Host` can come from three places (lowest priority first):
- Config File
- Environment
- Command Line

And this only uses env. IIRC either the Client or the factory has a GetHost method on it that you can use for this

borland (changes requested):
please look into the check failure and env.Gethost

pkg/cmd/project/view/view.go
benPearce1: I believe that this will be fixed when the update to `go-octopusdeploy` is fixed. I was using the following line during development to use an updated version:

`replace github.com/OctopusDeploy/go-octopusdeploy/v2 => ../go-octopusdeploy/`

pkg/cmd/space/view/view.go
benPearce1: `Factory.GetCurrentHost()` looks like it will do the trick 👍

pkg/cmd/project/view/view.go
borland: Oh right, makes sense.

I find it helpful in these situations to leave a comment on the PR next to the failing code and say something like "This will be resolved when https://github.com/OctopusDeploy/go-octopusdeploy/pull/145 merges"

benPearce1: I called it out, but could have been more explicit, plus I didn't know which specific bit was going to be impacted, still a bit green

borland (approved):

PR #111, opened 2022-10-13
chore: code refactor and testing changes

domenicsim1:


benPearce1 (approved):
this looks good, feel free to merge to the other branch and I can fix up any issues

pkg/cmd/projectgroup/create/create.go
```diff
@@ -153,3 +128,19 @@ func PromptMissing(opts *CreateOptions) error {
 
 	return nil
 }
+
+func AskName(ask question.Asker, messagePrefix string, resourceDescription string, value *string) error {
```
slewis74, line 132: we talked about having this in a shared package, will that be in a separate PR?

slewis74 (approved):
Very nice. Couple of whitespace fixes as suggestions

pkg/cmd/projectgroup/create/create.go
domenicsim1: My goal here we to get everything ready @benPearce1 so he could write tests and finish off project create, as a result of forgot to move it 😅 . I will move it, and fix the small nits but there will be another PR after that updates all the commands to use the common shared questions.

PR #112, opened 2022-10-13
feat: project delete

benPearce1:
[sc-26927]

delete via name
delete via slug
delete with confirm

image

### DIFF
--- a/pkg/cmd/account/delete/delete.go
+++ b/pkg/cmd/account/delete/delete.go
@@ -15,6 +15,7 @@ import (
 )
 
 func NewCmdDelete(f factory.Factory) *cobra.Command {
+	var skipConfirmation bool
 	cmd := &cobra.Command{
 		Use:     "delete {<name> | <id>}",
 		Short:   "Delete an account in an instance of Octopus Deploy",
@@ -30,11 +31,6 @@ func NewCmdDelete(f factory.Factory) *cobra.Command {
 			}
 
 			itemIDOrName := args[0]
-
-			skipConfirmation, err := cmd.Flags().GetBool("confirm")
-			if err != nil {
-				return err
-			}
 
 			client, err := f.GetSpacedClient()
 			if err != nil {
@@ -70,8 +66,8 @@ func NewCmdDelete(f factory.Factory) *cobra.Command {
 			return delete(client, itemToDelete)
 		},
 	}
-	// TODO confirm might want to be a global flag?
-	cmd.Flags().BoolP("confirm", "y", false, "Don't ask for confirmation before deleting the account.")
+
+	question.RegisterDeleteFlag(&skipConfirmation, cmd, "account")
 
 	return cmd
 }
--- a/pkg/cmd/account/delete/delete.go
+++ b/pkg/cmd/account/delete/delete.go
@@ -67,7 +67,7 @@ func NewCmdDelete(f factory.Factory) *cobra.Command {
 		},
 	}
 
-	question.RegisterDeleteFlag(&skipConfirmation, cmd, "account")
+	question.RegisterDeleteFlag(cmd, &skipConfirmation, "account")
 
 	return cmd
 }
--- a/pkg/cmd/project/delete/delete.go
+++ b/pkg/cmd/project/delete/delete.go
@@ -1,0 +1,1 @@
+package delete
--- a/pkg/cmd/project/delete/delete.go
+++ b/pkg/cmd/project/delete/delete.go
@@ -8,16 +8,13 @@ import (
 	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/client"
 	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/projects"
 	"github.com/spf13/cobra"
-	"io"
 )
 
 type DeleteOptions struct {
 	Client   *client.Client
 	Ask      question.Asker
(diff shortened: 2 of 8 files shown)


pkg/cmd/account/delete/delete.go
```diff
@@ -70,8 +66,8 @@ func NewCmdDelete(f factory.Factory) *cobra.Command {
 			return delete(client, itemToDelete)
 		},
 	}
-	// TODO confirm might want to be a global flag?
-	cmd.Flags().BoolP("confirm", "y", false, "Don't ask for confirmation before deleting the account.")
+
+	question.RegisterDeleteFlag(&skipConfirmation, cmd, "account")
```
borland, line 70: I feel like this would read better if `cmd` was first.

This is very subjective and opinion based. Imagine if it were C# and we were writing an extension method, it would more likely go on the cobra command than it would on the skipConfirmation variable. 
Take it or leave it 🤷

pkg/cmd/project/delete/delete.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/client"
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/projects"
+	"github.com/spf13/cobra"
+	"io"
+)
+
+type DeleteOptions struct {
+	Client   *client.Client
+	Ask      question.Asker
+	Host     string
```
borland, line 17: Host doesn't appear to be used at all?

pkg/cmd/project/delete/delete.go
```diff
@@ -0,0 +1,101 @@
...
+	Writer   io.Writer
```
borland, line 18: Writer doesn't appear to be used?

pkg/cmd/project/delete/delete.go
```diff
@@ -0,0 +1,101 @@
...
+	NoPrompt bool
+	idOrName string
```
borland, line 20: Is there any reason why `idOrName` is a private variable where the others are public?

(struct variables that start with lowercase letters are package-private in go. Uppercase are public.)

pkg/question/input.go
```diff
@@ -2,11 +2,28 @@ package question
...
+}
+
+func NewDeleteFlags() *DeleteFlags {
+	return &DeleteFlags{
+		Confirm: flag.New[bool](FlagConfirm, false),
+	}
+}
+
+func RegisterDeleteFlag(value *bool, cmd *cobra.Command, resourceDescription string) {
+	cmd.Flags().BoolVarP(value, FlagConfirm, "c", false, fmt.Sprintf("Don't ask for confirmation before deleting the %s.", resourceDescription))
```
borland, line 24: Should confirm have a short flag of `-c` here?
Short flags are designed to make it quick and easy to do something (e.g. -p for project because you specify it all the time). "short and easy" doesn't feel quite right for deleting stuff tho?
@jbristowe is the arbiter of this kind of thing, suggest getting his opinion if you haven't already.

borland (changes requested):

pkg/question/input.go
benPearce1: actually the shorthand was `y`, not `c`

pkg/cmd/project/delete/delete.go
benPearce1: no reason, I might have copied it from somewhere else, still getting my muscle memory in tune. Fixed.

pkg/cmd/project/delete/delete.go
benPearce1: removed unused variables

domenicsim1 (approved):
Just the changes @borland has requested and some questions but happy to move the question changes to a future PR as they depend on changes we have made with `project create`.
Overall, good stuff!

pkg/cmd/project/delete/delete.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/OctopusDeploy/cli/pkg/factory"
+	"github.com/OctopusDeploy/cli/pkg/question"
+	"github.com/OctopusDeploy/cli/pkg/question/selectors"
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/client"
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/projects"
+	"github.com/spf13/cobra"
+	"io"
+)
+
+type DeleteOptions struct {
```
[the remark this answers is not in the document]
benPearce1: 👍 I have removed the unused ones, can fix things up after `project create` goes in

pkg/cmd/project/delete/delete.go
```diff
@@ -0,0 +1,101 @@
...
+		return err
+	}
+
+	if opts.DeleteFlags.Confirm.Value {
+		return delete(opts.Client, itemToDelete)
+	} else {
+		return question.DeleteWithConfirmation(opts.Ask, "project", itemToDelete.Name, itemToDelete.ID, func() error {
+			return delete(opts.Client, itemToDelete)
+		})
+	}
```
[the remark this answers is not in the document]
benPearce1: There should be no reason for delete to have dependencies or be a dependency. Cascade deletes should be handled by the API.

borland (approved):
LGTM

PR #122, opened 2022-10-19
feat: project connect-tenant

benPearce1:
[sc-27715]

Simply a project command wrapping the `tenant connect` command
 
image

image

image

### DIFF
--- a/pkg/cmd/project/connect-tenant/connect-tenant.go
+++ b/pkg/cmd/project/connect-tenant/connect-tenant.go
@@ -18,7 +18,7 @@ func NewCmdConnect(f factory.Factory) *cobra.Command {
 		Long:  "Connect a tenant to a project in Octopus Deploy",
 		Example: fmt.Sprintf(heredoc.Doc(`
 			$ %s project connect
-			$ %s project connect --project "Deploy web site" --environment "Production"
+			$ %s project connect --tenant "Bobs Wood Shop" --project "Deploy web site" --environment "Production"
 		`), constants.ExecutableName, constants.ExecutableName),
 		RunE: func(c *cobra.Command, args []string) error {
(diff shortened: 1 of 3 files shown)


jbristowe:
I think we should name this `octopus project connect`. Conversely, we could also have a `octopus project disconnect`.

pkg/cmd/project/connect-tenant/connect-tenant.go
```diff
@@ -0,0 +1,32 @@
...
+
+func NewCmdConnect(f factory.Factory) *cobra.Command {
+	connectFlags := connectTenant.NewConnectFlags()
+	cmd := &cobra.Command{
+		Use:   "connect",
+		Short: "Connect a tenant to a project in Octopus Deploy",
+		Long:  "Connect a tenant to a project in Octopus Deploy",
+		Example: fmt.Sprintf(heredoc.Doc(`
+			$ %s project connect
+			$ %s project connect --project "Deploy web site" --environment "Production"
```
jbristowe, line 21: Is this right? Should it be `--tenant`?

domenicsim1: Yes, this is correct but I think we should have an example that shows specifying both project and tenant.

domenicsim1 (approved):

pkg/cmd/project/connect-tenant/connect-tenant.go
jbristowe: This should read, `%s project connect --tenant "Test Tenant" --environment "Production"`

benPearce1: updated both commands to include `--tenant` in the example

PR #133, opened 2022-11-06
feat: tenant create, delete, tag

domenicsim1:


pkg/cmd/project/delete/delete.go
```diff
@@ -37,6 +37,10 @@ func NewCmdList(f factory.Factory) *cobra.Command {
 				return err
 			}
 
+			if len(args) == 0 {
```
slewis74, line 40: I'm not clear on what this block is about. Is it actually required?

pkg/cmd/tenant/delete/delete.go
```diff
@@ -0,0 +1,109 @@
...
+			$ %s tenant delete
+			$ %s tenant rm
+		`), constants.ExecutableName, constants.ExecutableName),
+		RunE: func(_ *cobra.Command, args []string) error {
+			client, err := f.GetSpacedClient()
+			if err != nil {
+				return err
+			}
+
+			if len(args) == 0 {
```
slewis74, line 41: per previous comment

pkg/cmd/tenant/delete/delete.go
```diff
@@ -0,0 +1,109 @@
...
+	} else {
+		return question.DeleteWithConfirmation(opts.Ask, "tenant", itemToDelete.Name, itemToDelete.ID, func() error {
+			return delete(opts.Client, itemToDelete)
+		})
+	}
+}
+
+func PromptMissing(opts *DeleteOptions) error {
+	if opts.IdOrName == "" {
+		existingTenants, err := opts.Client.Tenants.GetAll()
```
slewis74, line 89: this can be an expensive call in some customer instances. I wonder if we should be using the paged results and a partial name filter?

slewis74 (approved):
Looking good. A couple of minor questions inline for consideration.

pkg/cmd/project/delete/delete.go
domenicsim1: yes its required because under this we try to access `args[0]` and if you just run `octopus project delete` go will panic with an index out of bounds when trying to do so. There for we will never be able to reach the no prompt mode where we select a project to delete in interactive mode.

PR #135, opened 2022-11-07
feat: tenant view

domenicsim1:


pkg/cmd/tenant/view/view.go
```diff
@@ -0,0 +1,112 @@
...
+		return err
+	}
+
+	fmt.Fprintf(opts.out, "%s %s\n", output.Bold(tenant.Name), output.Dimf("(%s)", tenant.ID))
+
+	if len(tenant.TenantTags) > 0 {
+		fmt.Fprintf(opts.out, "Tags: ")
+	}
+	for _, tag := range tenant.TenantTags {
+		fmt.Fprintf(opts.out, "%s, ", tag)
```
slewis74, line 90: It would be awesome if we could use the tag's colour as the text background colour here. Is that workable in the terminal or likely to cause us issues?

slewis74 (approved):
Looks good. Would love to see a screenshot in the conversation for these PRs, to help verify exactly what the layout will look like.

pkg/cmd/tenant/view/view.go
slewis74: Outcome from discussion: Not viable, at least not atm

PR #149, opened 2022-11-21
feat: documentation generation

jbristowe:


cmd/gen-docs/main.go
```diff
@@ -0,0 +1,134 @@
...
+	clientFactory := apiclient.NewStubClientFactory()
+	s := spinner.New(spinner.CharSets[11], 100*time.Millisecond, spinner.WithColor("cyan"))
+	buildVersion := strings.TrimSpace(version.Version)
+	f := factory.New(clientFactory, askProvider, s, buildVersion)
+
+	cmd := root.NewCmdRoot(f, clientFactory, askProvider)
+	cmd.DisableAutoGenTag = true
+	cmd.InitDefaultHelpCmd()
+
+	if err := os.MkdirAll(*dir, 0755); err != nil {
```
borland, line 66: is `0755` safe here?
(I'm not criticizing; I genuinely don't know -- this seems like a bit of a hard question. If we're doing something like copying github then a comment explaining that would be helpful.)

pkg/cmd/account/delete/delete.go
```diff
@@ -18,13 +18,13 @@ func NewCmdDelete(f factory.Factory) *cobra.Command {
...
+		Short:   "Delete an account",
+		Long:    "Delete an account in Octopus Deploy",
 		Aliases: []string{"del", "rm", "remove"},
-		Example: fmt.Sprintf(heredoc.Doc(`
-			$ %s account delete
-			$ %s account rm
-		`), constants.ExecutableName, constants.ExecutableName),
+		Example: heredoc.Docf(`
+			$ %[1]s account delete
+			$ %[1]s account rm
```
borland, line 26: I HAD NO IDEA YOU COULD DO THIS. Thanks :smile:

borland (approved):

cmd/gen-docs/main.go
jbristowe: Adjusted to `0644` 😉

PR #152, opened 2022-11-23
chore: tool info added to user agent string

benPearce1:
[sc-31593]

### DIFF
--- a/pkg/apiclient/client_factory.go
+++ b/pkg/apiclient/client_factory.go
@@ -21,11 +21,11 @@ import (
 type ClientFactory interface {
 	// GetSpacedClient returns an Octopus api Client instance which is bound to the Space
 	// specified in the OCTOPUS_SPACE environment variable, or the command line. It should be the default
-	GetSpacedClient() (*octopusApiClient.Client, error)
+	GetSpacedClient(requester Requester) (*octopusApiClient.Client, error)
 
 	// GetSystemClient returns an Octopus api Client instance which isn't bound to any Space.
 	// Use it for things that live outside of a space, such as Teams, or Spaces themselves
-	GetSystemClient() (*octopusApiClient.Client, error)
+	GetSystemClient(requester Requester) (*octopusApiClient.Client, error)
 
 	// GetActiveSpace returns the currently selected space.
 	// Note this is lazily populated when you call GetSpacedClient;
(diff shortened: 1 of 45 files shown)


pkg/apiclient/requester.go
```diff
@@ -0,0 +1,62 @@
...
+func NewRequester(c *cobra.Command) *RequesterContext {
+	return &RequesterContext{
+		cmd: c,
+	}
+}
+
+func (r *FakeRequesterContext) GetRequester() string { return "octopus/0.0.0" }
+
+func (r *RequesterContext) GetRequester() string {
+	version := "0.0.0"
```
borland, line 31: The easier way to get the CLI version is to use `version.Version` - it picks up the thing from version.txt so its basically a compile time constant (see version.go in the CLI source root folder)

pkg/apiclient/requester.go
```diff
@@ -0,0 +1,62 @@
...
+	for parentCmd != nil {
+		name := parentCmd.Name()
+		if name == constants.ExecutableName && version != "" {
+			rootCmd = fmt.Sprintf("%s/%s", name, version)
+		} else {
+			commands = append([]string{name}, commands...)
+		}
+		parentCmd = parentCmd.Parent()
+	}
+	return fmt.Sprintf("%s (%s)", rootCmd, strings.Join(commands, ";"))
```
borland, line 61: Not sure if I'm interpreting this correctly, it looks like the user agent will be something like `octopus/1.0.0 (release;create)`

Is that right? If so, I'm just curious as to what the benefit is? It seems like it will make server-side parsing of the user agent a bunch harder

benPearce1: this part is appended to the user agent generated by the go client, which you can see in the screenshots in the PR body

jbristowe: We are restricted by RFC7230 in terms of the delimiters we use. In this case, we use parentheses (comments), which is allowed by the specification.

pkg/apiclient/requester.go
benPearce1: updated to read from the`version.txt` file via `version.Version`

pkg/apiclient/client_factory.go
```diff
@@ -21,11 +21,11 @@ import (
 type ClientFactory interface {
 	// GetSpacedClient returns an Octopus api Client instance which is bound to the Space
 	// specified in the OCTOPUS_SPACE environment variable, or the command line. It should be the default
-	GetSpacedClient() (*octopusApiClient.Client, error)
+	GetSpacedClient(requester Requester) (*octopusApiClient.Client, error)
```
borland, line 24: Given that `Requester` is basically just `func () string` perhaps the code would be simpler and easier to understand if the signature changed to `GetSpacedClient(userAgent func() string)`? When I saw the 'Requester' interface I assumed some complex networking behaviour was going on, and I had to spend the time to figure it out. `func() string` would have made this more obvious.

You wouldn't need the ceremony with the fakeRequesterContext either

borland (commented):
Suggested an API simplification; Will leave it to you to discuss with the rest of integrations and make or reject the suggestion as you please

domenicsim1 (approved):

pkg/apiclient/client_factory.go
benPearce1: Thanks @borland, agreed that this could be an inline func, but I am going to leave it as it for the time being.

Issue #131, opened 2022-10-31
octopus space delete --confirm does not delete the space


borland:
Repro:

- Create a space
- run `octopus space delete "SpaceNameHere" --confirm`
- Expected: the space to get deleted
- Observed: the cli just exits with no output or warnings and the space is not deleted

I also tried "--confirm y", "--confirm=yes", "--confirm yes" and other variations like that.

Deleting spaces does work if you don't supply `--confirm` and go through the interactive confirmation.

Root cause is this code here:

https://github.com/OctopusDeploy/cli/blob/18b36f680e56a9d91d31658b208382da57f2d2c2/pkg/cmd/space/delete/delete.go#L51

The `delete` function call only runs if it's inside the question DeleteWithConfirmation callback, where it should run on both sides of the `if !alreadyConfirmed` branch

benPearce1:
@borland it appears that https://github.com/OctopusDeploy/cli/pull/169 fixes this issue, is that correct?

PR #169, opened 2022-12-08
fix: space delete --confirm 

borland:
Resolves issue 131
https://github.com/OctopusDeploy/cli/issues/131

### DIFF
--- a/pkg/cmd/space/delete/delete.go
+++ b/pkg/cmd/space/delete/delete.go
@@ -14,6 +14,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
+// TODO this command has no tests and doesn't follow the same patterns as our newer commands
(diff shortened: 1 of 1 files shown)


pkg/cmd/space/delete/delete.go
```diff
@@ -32,26 +33,28 @@ func NewCmdDelete(f factory.Factory) *cobra.Command {
 
 			itemIDOrName := args[0]
 
-			client, err := f.GetSystemClient(apiclient.NewRequester(cmd))
+			octopus, err := f.GetSystemClient(apiclient.NewRequester(cmd))
```
borland, line 36: rename to avoid name clash with 'client' package. Also we tend to call this 'octopus' everywhere else too

domenicsim1 (approved):

PR #172, opened 2022-12-13
fix: no prompt if only a single channel

benPearce1:
there is a redundant prompt if there is only a single channel

before:
image

after:
image


benPearce1:
@jbristowe 
> Would it be possible to inform the user, "Selecting channel, `Default` since it is the only channel."?

image

jbristowe (approved):

PR #174, opened 2022-12-14
fix: expand tenant view output

benPearce1:
This change is to support filtering tenants by tag using `jq`

json before:
image

json after:
image

* note: the change to pull all fields from the api means that we end up with the default json serialisation on the client model. In the case of the Tenants, the `TenantTags` field is marked as `omitempty` so an empty array ends up being left out. Previously we were ending up with the default values of an empty array in the json, making the contract more stable.

image

table before:
image

table after:
image

### DIFF
--- a/pkg/cmd/tenant/list/list.go
+++ b/pkg/cmd/tenant/list/list.go
@@ -77,6 +77,7 @@ func listRun(cmd *cobra.Command, f factory.Factory) error {
 				projectEnvironments = append(projectEnvironments, ProjectEnvironment{Project: projectEntity, Environments: environments})
 			}
 
+			t.Links = nil // ensure the links collection is not serialised
 			return TenantAsJson{
 				Tenant:              t,
 				ProjectEnvironments: projectEnvironments,
--- a/pkg/cmd/tenant/list/list.go
+++ b/pkg/cmd/tenant/list/list.go
@@ -6,14 +6,22 @@ import (
 	"github.com/OctopusDeploy/cli/pkg/constants"
 	"github.com/OctopusDeploy/cli/pkg/factory"
 	"github.com/OctopusDeploy/cli/pkg/output"
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/client"
 	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/tenants"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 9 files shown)


pkg/cmd/tenant/list/list.go
```diff
@@ -45,22 +53,78 @@ func listRun(cmd *cobra.Command, f factory.Factory) error {
...
+	for _, k := range keys {
+		entities = append(entities, output.IdAndName{Id: k, Name: lookup[k]})
+	}
+
+	return entities, nil
+}
+
+func getEnvironmentMap(client *client.Client) (map[string]string, error) {
+	environmentMap := make(map[string]string)
+	allEnvs, err := client.Environments.GetAll()
```
jbristowe, line 110: We should avoid `GetAll`. It does not scale. Is there any way of getting this information from another API?

pkg/cmd/tenant/list/list.go
```diff
@@ -45,22 +53,78 @@ func listRun(cmd *cobra.Command, f factory.Factory) error {
...
+	}
+	for _, e := range allEnvs {
+		environmentMap[e.GetID()] = e.GetName()
+	}
+	return environmentMap, nil
+}
+
+func getProjectMap(client *client.Client) (map[string]string, error) {
+	projectMap := make(map[string]string)
+	allEnvs, err := client.Projects.GetAll()
```
jbristowe, line 122: We should avoid `GetAll`. It does not scale. Is there any way of getting this information from another API?

pkg/cmd/tenant/list/list.go
benPearce1: fair point, I can modify it to find all the ids it needs and call the query endpoint with a list

jbristowe: Just out of curiosity, which endpoint can we use?

benPearce1: we can filter by a list of id values
`api/Spaces-1/environments?ids=Environments-1,Environments-282`

pkg/cmd/tenant/list/list.go
domenicsim1: agreed, in the api for all resources when you do `client.Projects.Get()` on the returned struct you can use the `.GetAllPages()` function to get all.

jbristowe (approved):

PR #182, opened 2022-12-16
chore: syntax error in example

benPearce1:


examples.md
```diff
@@ -115,4 +115,4 @@ octopus project convert --project 'Project 54' \
   --no-prompt
 ```
 
-An existing project can be converted to Config As Code using the `convert` command
\ No newline at end of file
+An existing project can be converted to Config As Code using the `convert` command
```
benPearce1, line 118: 🤷 not sure where that came from, must be whitspace

jbristowe (approved):

PR #189, opened 2022-12-20
chore: integration test on push

domenicsim1:


jbristowe (approved):

.github/workflows/integration-test.yml
```diff
@@ -5,6 +5,7 @@ on:
   schedule:
```
benPearce1, line 5: if we are running this on a branch push, including main, then we can probably drop the schedule

PR #191, opened 2022-12-22
Fixed the OCTOPUS_URL environment variable name

slewis74:
All of our GitHubAction tooling etc uses `OCTOPUS_URL` for the environment variable name that gives us the base url for the Octopus API.

This PR brings the Go CLI into line with that naming.


benPearce1 (approved):
LGTM, good to fix this before 1.0

PR #194, opened 2022-12-23
Fixes for Go Releaser, moving it to Ocotpus' v3 GitHub Actions

slewis74:
Go releaser workflow wasn't updated when the environment variable name was corrected, which meant that the use of the v2 GitHub Actions caused a failure.
This PR upgrades to v3 of the GitHub Actions

### DIFF
--- a/.github/workflows/go-releaser.yml
+++ b/.github/workflows/go-releaser.yml
@@ -219,10 +219,6 @@ jobs:
     - name: Copy publish-apt.sh and publish-rpm.sh
       run: cp linux-package-feeds/source/publish-*.sh artifacts/
 
-    - name: Create combined zipfile package for Octopus Deploy
-      working-directory: artifacts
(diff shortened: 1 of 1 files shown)


jbristowe (approved):

domenicsim1 (approved):

hnrkndrssn (approved):

.github/workflows/go-releaser.yml
```diff
@@ -223,31 +223,33 @@ jobs:
       working-directory: artifacts
       run: zip -r octopus-cli.${{ needs.goreleaser.outputs.version }}.zip .
```
jbristowe, line 225: Shouldn't we remove this?

slewis74: yep, forgot to delete it once I'd copied the behaviour 🙇‍♂️

PR #201, opened 2023-01-03
feat/pack

domenicsim1:


pkg/cmd/package/support/pack.go
```diff
@@ -0,0 +1,276 @@
...
+		}
+	}
+
+	if !opts.Verbose.Value {
+		result, err := question.SelectMap(opts.Ask, "Verbose", []bool{true, false}, func(b bool) string {
+			if b {
+				return "True"
+			} else {
+				return "False"
+			}
```
benPearce1, line 134: `verbose` and `overwrite` prompts could be `survey.Confirm`

pkg/cmd/package/zip/create/create.go
```diff
@@ -0,0 +1,76 @@
...
+
+	if opts.Id.Value == "" {
+		return errors.New("must supply a package ID")
+	}
+	applyDefaultsToUnspecifiedOptions(opts)
+
+	pack.VerboseOut(opts.Verbose.Value, "Packing \"%s\" version \"%s\"...\n", opts.Id.Value, opts.Version.Value)
+
+	outFilePath := pack.BuildOutFileName("zip", opts.Id.Value, opts.Version.Value)
+	return pack.BuildPackage(opts, outFilePath)
```
benPearce1, line 57: automation cmd output is missing

pkg/cmd/package/nuget/create/create.go
```diff
@@ -0,0 +1,315 @@
...
+
+func createRun(opts *NuPkgCreateOptions) error {
+	if !opts.NoPrompt {
+		if err := pack.PackageCreatePromptMissing(opts.PackageCreateOptions); err != nil {
+			return err
+		}
+		if err := PromptMissing(opts); err != nil {
+			return err
+		}
+
```
benPearce1, line 99: whitespace

pkg/cmd/package/support/pack.go
```diff
@@ -0,0 +1,276 @@
...
+		fmt.Printf(messageTemplate, messageArgs...)
+	}
+}
+
+func BuildTimestampSemVer(dateTime time.Time) string {
+	endSegment := dateTime.Hour()*10000 + dateTime.Minute()*100 + dateTime.Second()
+	return fmt.Sprintf("%d.%d.%d.%d", dateTime.Year(), dateTime.Month(), dateTime.Day(), endSegment)
+}
+
+func BuildOutFileName(packageType string, id, version string) string {
```
benPearce1, line 170: inconsistent layout
```suggestion
func BuildOutFileName(packageType, id, version string) string {
```

pkg/cmd/package/support/pack.go
```diff
@@ -0,0 +1,276 @@
...
+			return err
+		}
+	}
+
+	if len(opts.Include.Value) == 0 {
+		for {
+			var pattern string
+			if err := opts.Ask(&survey.Input{
+				Message: "Include patterns",
+				Help:    "Add a file pattern to include, relative to the base path e.g. /bin/*.dll; defaults to \"**\".",
```
benPearce1, line 117: could have the default value of `**\.` in the prompt
also the default could be moved to a const

pkg/cmd/package/support/pack.go
```diff
@@ -0,0 +1,276 @@
...
+		}, &opts.BasePath.Value); err != nil {
+			return err
+		}
+	}
+
+	if opts.OutFolder.Value == "" {
+		if err := opts.Ask(&survey.Input{
+			Message: "Out Folder",
+			Help:    "Folder into which the zip file will be written; defaults to the current directory",
+		}, &opts.OutFolder.Value); err != nil {
```
benPearce1, line 107: set the default to '.'

pkg/cmd/package/support/pack.go
```diff
@@ -0,0 +1,276 @@
...
+		}, &opts.Version.Value); err != nil {
+			return err
+		}
+	}
+
+	if opts.BasePath.Value == "" {
+		if err := opts.Ask(&survey.Input{
+			Message: "Base Path",
+			Help:    "Root folder containing the contents to zip; defaults to current directory.",
+		}, &opts.BasePath.Value); err != nil {
```
benPearce1, line 98: set the default to `.`

pkg/cmd/package/support/pack.go
```diff
@@ -0,0 +1,276 @@
...
+			survey.MinLength(1),
+		))); err != nil {
+			return err
+		}
+	}
+
+	if opts.Version.Value == "" {
+		if err := opts.Ask(&survey.Input{
+			Message: "Version",
+			Help:    "The version of the package; must be a valid SemVer; defaults to a timestamp-based version.",
```
benPearce1, line 88: could precalculate the default timestamped version and show it in the prompt, this might not be a great idea but something to consider

pkg/cmd/package/nuget/create/create.go
```diff
@@ -0,0 +1,315 @@
...
+		}, &opts.Title.Value); err != nil {
+			return err
+		}
+	}
+
+	if opts.Description.Value == "" {
+		if err := opts.Ask(&survey.Input{
+			Message: "Nuspec description",
+			Help:    "The description to include in the Nuspec file; defaults to \"A deployment package created from files on disk.\".",
+		}, &opts.Description.Value); err != nil {
```
benPearce1, line 186: set the `Default` property to the default value

benPearce1 (changes requested):
several comments for changes

pkg/cmd/package/support/pack.go
domenicsim1: this one can not have a default in survey because of the loop it's in. If you don't provide a value its will set it to "**" making the loop never end, also makes it impossible to have "**" after providing other patterns.

benPearce1: good point

pkg/cmd/package/support/pack.go
domenicsim1: default is actually never set, removed text.

pkg/cmd/package/nuget/create/create.go
domenicsim1: should not set this one. Removed the text as it was never set. If we did set this as default nuspec file will try to generate all the time when it does not need too.

benPearce1: Agreed, but we could set it an author has been provided, since author is mandatory, but not important

go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/AlecAivazis/survey/v2 v2.3.6
 	github.com/MakeNowJust/heredoc/v2 v2.0.1
 	github.com/OctopusDeploy/go-octopusdeploy/v2 v2.20.0
+	github.com/bmatcuk/doublestar/v4 v4.4.0
```
borland, line 9: Nice. I didn't realise that go's builtin glob didn't suport **

pkg/cmd/package/nuget/create/create.go
```diff
@@ -0,0 +1,305 @@
...
+func NewCmdCreate(f factory.Factory) *cobra.Command {
+	createFlags := NewNuPkgCreateFlags()
+	packFlags := pack.NewPackageCreateFlags()
+
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "Create nuget",
+		Long:  "Create nuget package",
+		Example: heredoc.Docf(`
+			$ %[1]s project nuget create --id SomePackage --version 1.0.0
```
borland, line 62: "package nuget create"?

pkg/cmd/package/nuget/create/create.go
domenicsim1: agreed, I will also change the prompts so it only asks the other nuspec related question if you provide an author.

test/testutil/testutil.go
```diff
@@ -105,3 +107,32 @@ func ParseJsonStrict[T any](input io.Reader) (T, error) {
 	decoder.DisallowUnknownFields()
 	return parsedStdout, decoder.Decode(&parsedStdout)
 }
+
+// CaptureConsoleOutput borrows from `github.com/zenizh/go-capturer`. This implementation captures both stdout
+// and stderr. Consider adding the go-capturer module if more granularity is needed
+func CaptureConsoleOutput(f func()) string {
```
borland, line 113: I think there's already something very similar to this in the integration test util file; I used it when we launch the CLI as an external process and want to capture stdout/stderr

pkg/cmd/package/support/pack.go
```diff
@@ -0,0 +1,270 @@
...
+	"strings"
+	"time"
+)
+
+const (
+	FlagId        = "id"
+	FlagVersion   = "version"
+	FlagBasePath  = "base-path"
+	FlagOutFolder = "out-folder"
+	FlagInclude   = "include"
```
borland, line 26: Upon seeing an 'include' parameter, I went looking for an 'exclude' parameter, but there isn't one.

Can we add one please? It's super useful to be able to do things like `zip myarchive.zip * -x .git`, so I'd guess that would apply here too?

pkg/cmd/package/nuget/create/create.go
```diff
@@ -0,0 +1,314 @@
...
+				Message: message,
+				Help:    "Add an author to the package metadata.",
+			}, &author); err != nil {
+				return err
+			}
+			if author == "" {
+				break
+			}
+			opts.Author.Value = append(opts.Author.Value, author)
+			message = message + " (leave blank to continue)"
```
borland, line 168: This looks like it will keep asking for more and more authors until you finally enter an empty line, which is cool.

This string concatenation looks like the output will be 

```
Author:
Author (leave blank to continue):
Author (leave blank to continue) (leave blank to continue):
Author (leave blank to continue) (leave blank to continue) (leave blank to continue):
```

Also - it looks like author is optional? If it is, maybe we should put the "leave blank" bit up there on the first prompt as well.

pkg/cmd/package/nuget/create/create.go
```diff
@@ -0,0 +1,314 @@
...
+		message := "Author"
+		for {
+			var author string
+			if err := opts.Ask(&survey.Input{
+				Message: message,
+				Help:    "Add an author to the package metadata.",
+			}, &author); err != nil {
+				return err
+			}
+			if author == "" {
```
borland, line 164: suggest trimming the string before checking for empty

pkg/cmd/package/nuget/create/create.go
```diff
@@ -0,0 +1,314 @@
...
+				break
+			}
+			opts.Author.Value = append(opts.Author.Value, author)
+			message = message + " (leave blank to continue)"
+		}
+	}
+
+	if len(opts.Author.Value) > 0 {
```
borland, line 172: If I don't input an author, this then means I can't have a title/releasenotes/either. That seems a bit unexpected?

pkg/cmd/package/nuget/create/create.go
```diff
@@ -0,0 +1,314 @@
...
+		opts.OutFolder.Value = "."
+	}
+
+	if len(opts.Include.Value) == 0 {
+		opts.Include.Value = append(opts.Include.Value, "**")
+	}
+
+	if len(opts.Author.Value) > 0 {
+		if opts.Description.Value == "" {
+			opts.Description.Value = "A deployment package created from files on disk."
```
borland, line 236: This seems low-information. Maybe just have no description?

If we want to go with some generic placeholder, we could use it for a bit of shameless self-promotion and use "Created by Octopus CLI" as the description :smile:

pkg/cmd/package/nuget/create/create.go
```diff
@@ -0,0 +1,314 @@
...
+
+	filePath := filepath.Join(opts.BasePath.Value, opts.Id.Value+".nuspec")
+
+	var sb strings.Builder
+	sb.WriteString(`<?xml version="1.0" encoding="utf-8"?>` + "\n")
+	sb.WriteString(`<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">` + "\n")
+	sb.WriteString("  <metadata>\n")
+	sb.WriteString("    <id>" + opts.Id.Value + "</id>\n")
+	sb.WriteString("    <version>" + opts.Version.Value + "</version>\n")
+	sb.WriteString("    <description>" + opts.Description.Value + "</description>\n")
```
borland, line 295: Need to escape the string values here, otherwise you'll get a broken XML file if someone puts a `<` in their version/description/author/etc

pkg/cmd/package/nuget/create/create.go
```diff
@@ -0,0 +1,314 @@
...
+	sb.WriteString("    <authors>" + strings.Join(opts.Author.Value, ",") + "</authors>\n")
+	if releaseNotes != "" {
+		sb.WriteString("    <releaseNotes>" + releaseNotes + "</releaseNotes>\n")
```
borland, line 298: In line with escaping, maybe CDATA is appropriate for release notes here?

pkg/cmd/package/nuget/create/create.go
```diff
@@ -0,0 +1,314 @@
...
+			opts.BasePath,
+			opts.OutFolder,
+			opts.Include,
+			opts.Verbose,
+			opts.Overwrite,
+		)
+		fmt.Fprintf(opts.Writer, "\nAutomation Command: %s\n", autoCmd)
+	}
+
+	if nuspecFilePath != "" {
```
borland, line 146: If something fails we will leave behind the generated nuspec file which may then break the next time they run the command. Maybe use a defer to clean it up?

pkg/cmd/package/support/pack.go
```diff
@@ -0,0 +1,270 @@
...
+		err := opts.Ask(&survey.Confirm{
+			Message: "Verbose",
+			Default: true,
+		}, &opts.Verbose.Value)
+		if err != nil {
+			return err
+		}
+	}
+
+	if !opts.Overwrite.Value {
```
borland, line 140: This seems like a niche/advanced thing - maybe don't bother asking it in interactive mode (they can always specify --overwrite if they want to)

pkg/cmd/package/support/pack.go
```diff
@@ -0,0 +1,270 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+func VerboseOut(isVerbose bool, messageTemplate string, messageArgs ...any) {
+	if isVerbose {
+		fmt.Printf(messageTemplate, messageArgs...)
```
borland, line 155: IIRC we want to run message printing using the cobra cmd.print functions or other redirected IO; if you just do bare printf then things like color escape codes are broken on windows and NO_COLOR doesn't work

pkg/cmd/package/support/pack_test.go
```diff
@@ -0,0 +1,69 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	return dir
+}
+
+// TODO Test and potentially remove manual clean-up when go version >= 1.20.0
+// cleanUpTemp is a temporary solution for windows to https://github.com/golang/go/issues/51442.
+func cleanUpTemp(tempDir string) {
```
borland, line 63: Is tempDir here the actual user temp dir, or some subdirectory inside it?

If you are doing os.RemoveAll on `C:\Users\Orion\AppData\Local\Temp` and not a sub folder, there's about 10000 files in there and some of them will be in use by other running applications; it's not good form to nuke them like that that.

If it's a subdir of our own construction, nothing to see here :smile:

pkg/cmd/package/zip/create/create.go
```diff
@@ -0,0 +1,93 @@
...
+
+func NewCmdCreate(f factory.Factory) *cobra.Command {
+	createFlags := pack.NewPackageCreateFlags()
+
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "Create zip",
+		Long:  "Create zip package",
+		Example: heredoc.Docf(`
+			$ %[1]s project zip create --id SomePackage --version 1.0.0
```
borland, line 24: package zip craete?

pkg/cmd/package/nuget/create/create.go
domenicsim1: author is required for a nuspec file, if you don't have an author you don't need a nuspec file. No point asking for the others.

pkg/cmd/package/nuget/create/create.go
domenicsim1: I like the shameless plug idea, will bring it up with the team. The reason this has a default is because it required for a nuspec file.

pkg/cmd/package/nuget/create/create.go
benPearce1: we might have this same issue in `project convert` too https://github.com/OctopusDeploy/cli/blob/main/pkg/cmd/project/convert/convert.go#L296

pkg/cmd/package/nuget/create/create.go
borland: Ah cool. Maybe a small comment would be useful?

benPearce1 (approved):

Issue #207, opened 2023-01-06
Select prompts error when there are no options


benPearce1:
List-select prompts that have no options will error when there are no options available

e.g error due to target space having no environments 
```
❯ octopus deployment-target cloud-region create --space test1
? Name cr1
please provide options to select from
```

PR #208, opened 2023-01-06
Fix: empty list errors

benPearce1:
fixes: https://github.com/OctopusDeploy/cli/issues/207

before:
image

after:
image

before:
image

after:
image

### DIFF
--- a/pkg/question/select.go
+++ b/pkg/question/select.go
@@ -2,6 +2,7 @@ package question
 
 import (
 	"fmt"
+	"github.com/OctopusDeploy/cli/pkg/util"
 
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/OctopusDeploy/cli/pkg/constants"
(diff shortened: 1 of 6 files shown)


pkg/surveyext/multiselectwithadd.go
```diff
@@ -265,12 +264,6 @@ func (m *MultiSelectWithAdd) Prompt(config *survey.PromptConfig) (interface{}, e
 		}
 	}
 
-	// if there are no options to render
-	if len(m.Options) == 0 {
-		// we failed
-		return "", errors.New("please provide options to select from")
-	}
-
```
benPearce1, line 273: this copy of the multi-select is specifically designed to allow adding new entries, an empty starting list should not be an error

pkg/question/select.go
```diff
@@ -2,13 +2,17 @@ package question
...
+	"github.com/OctopusDeploy/cli/pkg/util"
 
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/OctopusDeploy/cli/pkg/constants"
 	"github.com/OctopusDeploy/cli/pkg/surveyext"
 )
 
 func MultiSelectMap[T any](ask Asker, message string, items []T, getKey func(item T) string, required bool) ([]T, error) {
+	if util.Empty(items) {
+		return nil, fmt.Errorf("%s - no options available", message)
```
benPearce1, line 14: by erroring with the original prompt, the user gets the context of the error

jbristowe (approved):

PR #216, opened 2023-01-27
feat: project variables include

benPearce1:
feat: project variables exclude

interactive list of variable sets
image

selecting and updating
image

specify variable set 
image

variable set already included
image

result:
image

remove variable set
image

removing var set not currently included
image


domenicsim1 (approved):

benPearce1:
BEGIN_COMMIT_OVERRIDE
feat: project variables include
feat: project variables exclude
END_COMMIT_OVERRIDE

PR #255, opened 2023-07-27
missed terminating template if statement

benPearce1:
Based on this example https://github.com/goreleaser/goreleaser/blob/main/.goreleaser.yaml#L153


hnrkndrssn (approved):
:+1: sorry I missed this in the previous PR

PR #270, opened 2023-09-07
feat: Adds support for authenticating with access token

geofflamrock:
This PR adds support for authenticating using an access token as a bearer token in the `Authorization` header. Access tokens are obtained using OpenID Connect and by nature are short-lived. As such, storing access tokens in the cli configuration (`octopus config set`) is not supported, and are only read from the `OCTOPUS_ACCESS_TOKEN` environment variable. This environment variable will be set automatically when using the `OctopusDeploy/login` action which is currently only released internally.

Note: This PR requires https://github.com/OctopusDeploy/go-octopusdeploy/pull/201 to be completed and a new version of the `go-octopusdeploy` client to be released before it will build. Update: This PR is merged now.

Note: The OpenID Connect feature is currently under development and may not be available on all Octopus instances at the moment, so this PR doesn't make any changes to the readme for the CLI. Once the feature is released then a follow up PR will add documentation around this.

### DIFF
--- a/pkg/apiclient/client_factory.go
+++ b/pkg/apiclient/client_factory.go
@@ -57,10 +57,8 @@ type Client struct {
 
 	// the Server URL, obtained from OCTOPUS_URL
 	ApiUrl *url.URL
-	// the Octopus API Key, obtained from OCTOPUS_API_KEY
-	ApiKey string
-	// the Octopus Access Token, obtained from OCTOPUS_ACCESS_TOKEN
-	AccessToken string
+	// Credentials, obtained from OCTOPUS_API_KEY or OCTOPUS_ACCESS_TOKEN
+	Credentials octopusApiClient.ICredential
 	// the Octopus SpaceNameOrID to work within. Obtained from OCTOPUS_SPACE (TODO: or --space=XYZ on the command line??)
 	// Required for commands that need a space, but may be omitted for server-wide commands such as listing teams
 	SpaceNameOrID string
@@ -72,14 +70,16 @@ type Client struct {
 	Ask question.AskProvider
 }
 
-func NewClientFactory(httpClient *http.Client, host string, apiKey string, accessToken string, spaceNameOrID string, ask question.AskProvider) (ClientFactory, error) {
+func NewClientFactory(httpClient *http.Client, host string, credentials octopusApiClient.ICredential, spaceNameOrID string, ask question.AskProvider) (ClientFactory, error) {
 	// httpClient is allowed to be nil; it is passed through to the go-octopusdeploy library which falls back to a default httpClient
 	if host == "" {
 		return nil, cliErrors.NewArgumentNullOrEmptyError("host")
 	}
-	if apiKey == "" && accessToken == "" {
-		return nil, errors.New("one of apiKey or accessToken arguments are required")
+
+	if credentials == nil {
+		return nil, cliErrors.NewArgumentNullOrEmptyError("credentials")
 	}
+
 	// space is allowed to be blank, we will prompt for a space in interactive mode, or error if not
 	if ask == nil {
 		return nil, cliErrors.NewArgumentNullOrEmptyError("ask")
@@ -95,8 +95,7 @@ func NewClientFactory(httpClient *http.Client, host string, apiKey string, acces
 		SystemClient:      nil,
 		SpaceScopedClient: nil,
 		ApiUrl:            hostUrl,
-		ApiKey:            apiKey,
-		AccessToken:       accessToken,
+		Credentials:       credentials,
 		SpaceNameOrID:     spaceNameOrID,
 		ActiveSpace:       nil,
 		Ask:               ask,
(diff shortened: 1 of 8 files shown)


pkg/apiclient/client_factory.go
```diff
@@ -70,13 +72,13 @@ type Client struct {
 	Ask question.AskProvider
 }
 
-func NewClientFactory(httpClient *http.Client, host string, apiKey string, spaceNameOrID string, ask question.AskProvider) (ClientFactory, error) {
+func NewClientFactory(httpClient *http.Client, host string, apiKey string, accessToken string, spaceNameOrID string, ask question.AskProvider) (ClientFactory, error) {
```
johnsimons, line 75: I am not familiar with this code base, but to me, it feels weird to have both `apiKey` and `accessToken` in the initialiser function, aren't they mutually exclusive?

johnsimons (approved):
A small comment regarding the init function, but overall seems reasonable to me.
That said, this is the first time I am looking at this code base, so it may be a good idea to ping others who are more familiar with this code base.

pkg/apiclient/client_factory.go
geofflamrock: Technically they aren't mutually exclusive, if you supply both it will use both and the api key will win out on the Server. But yes for practical purposes you would only really use one or the other. In the client library we introduced a struct for this that's used in some function overloads (leaving the original ones in place to maintain backward compat), we could do the same here potentially. I'll have a look and see what the changes look like, it might cause the PR to get pretty large with changes to tests but that could be ok if it ends up more obvious and maintainable. Thanks.

geofflamrock: One other thought here is backward compatibility on these functions. I had made an assumption that we don't need to ensure this as we don't publish the packages for consumption elsewhere, I can't find the cli with a search https://pkg.go.dev/search?q=octopusdeploy, only our client and few other third-party packages. @benPearce1 do you know if this assumption is correct?

benPearce1: The only other place the `go-octopusdeploy` library is used is in our terrraform provider. I think it might be used by individuals as well, but not in other "products"

geofflamrock: The change in `go-octopusdeploy` are backward compatible, we left the original functions in there. I'm more thinking of the cli itself, I don't think it gets published as a consumable go module, only as a binary, is that right?

benPearce1: Sorry, misunderstood. That is correct.

geofflamrock: Thanks, I'll take a look at what the changes might look like to have a single credentials parameter instead of two like we do in `go-octopusdeploy`, might be worth it if the result is cleaner and easier to follow.

benPearce1: The change that we might prefer might look something like this in the client library

```
type ICredential interface {
	GetHeaderValue() (string, string)
}

type ApiKey struct {
	Value string
	ICredential
}

func (apikey *ApiKey) GetHeaderValue() (string, string) {
	return constants.ClientAPIKeyHTTPHeader, apikey.Value
}

func NewApiKey(apiKey string) (*ApiKey, error) {
	return &ApiKey{
		Value: apiKey,
	}, nil
}

type AccessToken struct {
	Value string
	ICredential
}

func (accessToken *AccessToken) GetHeaderValue() (string, string) {
	return "Authorization", fmt.Sprintf("Bearer %s", accessToken.Value)
}

func NewAccessToken(accessToken string) (*AccessToken, error) {
	return &AccessToken{
		Value: accessToken,
	}, nil
}

func getHeaders(apiCredentials ICredential, requestingTool string) map[string]string {
	headers := map[string]string{
		"User-Agent": api.GetUserAgentString(requestingTool),
	}

	headerKey, headerValue := apiCredentials.GetHeaderValue()
	headers[headerKey] = headerValue

	return headers
}
```

geofflamrock: That is certainly nicer than what we have at the moment, I'm very new to go and so went with the simplest option I could do at the time but this is a nicer approach. If we were to do this change in the client and then roll it to the cli, given we've shipped the new minor version of the client in its current form would it constitute a breaking change?

benPearce1 (changes requested):
I would like to see the `ApiCredentials` struct dropped in favour of an interface with an Api Key implementation and Access Token implementation. 
This is actually a change in the client that would impact this PR

pkg/apiclient/client_factory.go
benPearce1: In theory, the previous change was also a breaking change due to the new parameter.

geofflamrock: We only use it in new functions and kept the old ones around that passed through the correct values, so I didn't consider it to be a breaking change?

benPearce1: Cool, didn't see that. Then yes, this proposal would be a breaking change.

geofflamrock: We only shipped this a couple of days ago, would we need to consider a new major version with this change or can we just push a new minor? It would be nice to avoid a major version if possible.

geofflamrock: Actually it was only yesterday that we shipped it

benPearce1: These go projects aren't using strict semver, a minor version bump would be sufficient

geofflamrock: Ok great, I'll make the change and might send it your way for a review if thats ok

benPearce1: No problem

geofflamrock: Changes to the client have been made in https://github.com/OctopusDeploy/go-octopusdeploy/pull/202 and the equivalent changes made in this PR. It won't build just yet as it'll require the client changes to be released first. @benPearce1 do you mind also eyeballing the changes here to see if the usage feels good now please? If so I'll merge the client PR and get this one updated.

benPearce1 (approved):
👍

PR #291, opened 2023-11-14
feat: Add new login and logout commands

geofflamrock:
# Commands

This PR adds support for two new commands to the cli.

## `login`

This command supports logging in to an Octopus Server in multiple ways.

### OpenID Connect (OIDC)

This is for CI/automation scenarios outside of GitHub Actions, which we have a [login action](https://github.com/OctopusDeploy/login) for. 

It takes in `--server`, `--service-account-id` and `--id-token`, will perform the OIDC token exchange with the Octopus Server to obtain an access token, test the connection to the Server and configure the CLI ready for use.

OIDC

### API Key (interactive)

This will prompt the user for the server url, and if they want to create a new API key. If they choose to create a new key it will open up a browser at the API key screen for their profile where they can create the key and then paste it into the terminal. Once pasted in, the key will be tested by performing a query to the Server and the CLI configured ready for use.

API Key Interactive

### API Key (non-interactive)

This takes in `--server` and `--api-key`, will test the key against the Server and configure the CLI ready for use.

## `logout`

This command removes any configured Server, API key or Access Token, leaving the environment back to where it was at the start.

# Other changes

This PR also changes the prompt shown to the user when they don't have the CLI configured:
- If the CLI is in interactive mode, it will prompt them to use `octopus login` or configure the `OCTOPUS_URL`/`OCTOPUS_API_KEY` environment variables.

image

- If the CLI is in non-interactive mode (e.g. a CI server), it will prompt them to use the non-interactive versions of the login command for OIDC or API key.

image


pkg/cmd/config/set/set.go
```diff
@@ -118,3 +118,42 @@ func promptMissing(ask question.Asker, key string) (string, string, error) {
 
 	return key, value, nil
 }
+
+func SetConfig(key string, value string) error {
```
geofflamrock, line 122: TODO: This was copied from `setRun` up above, need to combine them back again

pkg/cmd/login/login.go
```diff
@@ -0,0 +1,385 @@
...
+	AccessToken string `json:"access_token"`
+	ExpiresIn   int32  `json:"expires_in"`
+}
+
+type TokenExchangeErrorResponse struct {
+	Error            string `json:"error"`
+	ErrorDescription string `json:"error_description"`
+}
+
+func loginRun(cmd *cobra.Command, f factory.Factory, isPromptEnabled bool, ask question.Asker, flags *LoginFlags) error {
```
geofflamrock, line 102: Need to add tests for this

pkg/cmd/root/help.go
```diff
@@ -66,7 +66,7 @@ func rootHelpFunc(cmd *cobra.Command, _ []string) {
 		helpEntries = append(helpEntries, helpEntry{"", longText})
 	}
 	if isRootCmd(cmd) && viper.GetBool(constants.ConfigShowOctopus) {
-		helpEntries = append(helpEntries, helpEntry{"", output.Bluef("%s", constants.OctopusLogo)})
+		helpEntries = append(helpEntries, helpEntry{"", output.Cyanf("%s", constants.OctopusLogo)})
```
geofflamrock, line 69: This was a deliberate change to make the colours of the Octopus logo printed match our branding better

pkg/constants/constants.go
```diff
@@ -53,23 +53,24 @@ const (
...
-      &%%%%%%%%%%%%%%%%%%%%%G
-       B%%%%%%%%%%%%%%%%%%%%#
-       P%%%%%%%%%%%%%%%%%%%%P
-     &P%%%%%%%%%%%%%%%%%%%%%%P#
-   #G%%%%P%%%%%%GP%%%%%%PP%%%%%PG&
-&G%%%PB&#P%%%%G   G%%%%P  &P%%P#&&
- &&     G%%%%#    &%%%%G    #P%%#
-       &%%%G&      G%%%B       &&
-        &#          BPB`
+const OctopusLogo = `                ####
```
geofflamrock, line 56: Using the logo as it was previously was causing some strange printing errors, I think because of all the `%` characters. I think it also looked a bit strange with all the different characters, so have changed it to just use `#`. I think it looks a bit nicer like this but definitely happy to discuss if there was a good reason for the characters that I'm not aware of or others disagree on the aesthetics.

pkg/apiclient/client_factory.go
```diff
@@ -40,6 +41,9 @@ type ClientFactory interface {
 
 	// GetHostUrl returns the current set API URL as a string
 	GetHostUrl() string
+
+	// GetHttpClient returns a raw http client which can be used to query Octopus
+	GetHttpClient() (*http.Client, error)
```
geofflamrock, line 46: We don't use `SystemClient` from the go-octopusdeploy library here because we need to send anonymous requests to the /.well-known/openid-configuration endpoint on the Octopus Server and then follow to the token endpoint. The system client assumes a base path of /api which doesn't apply here. 

So this function has been added to the factory to allow the login command to send the relevant requests but also allow us to mock these in tests.

pkg/cmd/config/list/list.go
```diff
@@ -39,6 +39,10 @@ func listRun(cmd *cobra.Command) error {
 		configFile.Set(constants.ConfigApiKey, "***")
 	}
 
+	if configFile.IsSet(constants.ConfigAccessToken) {
```
geofflamrock, line 42: We now set the access token directly into the config file (in non-interactive mode for OIDC), so we want to mask it's value if someone were to go asking for it.

pkg/config/provider.go
```diff
@@ -0,0 +1,59 @@
+package config
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/spf13/viper"
+)
+
+type IConfigProvider interface {
```
geofflamrock, line 10: This interface has been introduced to provide testable access to the config system for non-interactive usage e.g. when the login command is done we need to set the appropriate config.

At the moment this isn't used by the `octopus config get` or `octopus config set` commands as they have different requirements around prompting for missing keys and special handling of some values.

pkg/cmd/login/login.go
```diff
@@ -0,0 +1,404 @@
...
+				return nil, err
+			}
+
+			if createNewApiKey {
+				createApiKeyUrl := fmt.Sprintf("%s/app#/users/me/apiKeys", server)
+				createApiKeyLink := output.Cyan(createApiKeyUrl)
+				cmd.Printf("A web browser has been opened at %s. Please create an API key and paste it here. If no web browser is available or if the web browser fails to open, please use the --server and --api-key arguments directly e.g. octopus login --server %s --api-key API-MYAPIKEY.", createApiKeyLink, server)
+				cmd.Println()
+
+				err := browser.OpenURL(createApiKeyUrl)
```
geofflamrock, line 353: Would be nice to be able to test this behaviour somehow

test/testutil/fakefactory.go
```diff
@@ -2,27 +2,47 @@ package testutil
...
 
 func (f *FakeSpinner) Start() {}
 func (f *FakeSpinner) Stop()  {}
 
+type FakeConfigProvider struct {
+	config map[string]string
+}
+
+func (f *FakeConfigProvider) Get(key string) string {
+	return f.config[key]
```
geofflamrock, line 27: Hmmm wonder if we should check for `nil` here

zentron (approved):
This looks pretty good to me. Nice work getting stuck into this new codebase.

Issue #269, opened 2023-09-05
missing {-v|--version} flag


eckdanny:
New CLI missing this important flag:

```sh
# octopuscli
$ octo --version
  9.1.7
```

```sh
# octopus-cli
$ octopus --version
unknown flag: --version
```

Expectation from CLI users like me: _all CLIs_ support a `--help` and a `--version` flag, as the old client did

PR #298, opened 2023-11-29
Add version flag

thampton:
```
  ./octopus --version
```

Fixes https://github.com/OctopusDeploy/cli/issues/269

### DIFF
--- a/pkg/cmd/root/help.go
+++ b/pkg/cmd/root/help.go
@@ -6,6 +6,7 @@ import (
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


OzoneNZ:
@benPearce1  any chance of getting this reviewed and merged? We pack the `octopus` CLI into our custom worker-tools container image, and as part of that build process we really want to emit version info which we were able to do with the legacy CLI

benPearce1:
@OzoneNZ there is already a version command available, just in a slightly different format

```
❯ octopus version
2.1.0
```

Issue #297, opened 2023-11-29
octopus tenant variables list - unknown flag: --tenant


thampton:
## The bug

octopus tenant variables list --tenant "Bobs Wood Shop" returns "unknown flag: --tenant

Documentation:
```
Manage tenant variables in Octopus Deploy

USAGE
 octopus tenant variables <command> [flags]

CORE COMMANDS
 list:        List tenant variables
 update:      Update the value of a tenant variable

INHERITED FLAGS
 -h, --help                   Show help for a command
     --no-prompt              Disable prompting in interactive mode
 -f, --output-format string   Specify the output format for a command ("json", "table", or "basic") (default "table")
 -s, --space string           Specify the space for operations

EXAMPLE
 $ octopus tenant variables list --tenant "Bobs Wood Shop"
 $ octopus tenant variables view --name "DatabaseName" --tenant "Bobs Wood Shop"
```

Appears the format in this  https://github.com/OctopusDeploy/cli/pull/225 PR somewhat works but it doesn't match the documentation or examples.

```
% octopus tenant variables list "Blue"
? You have not specified a Space. Please select one: Pattern - Blue-Green
NAME  LABEL  TYPE  OWNER  ENVIRONMENT  VALUE  SENSITIVE  DEFAULT VALUE
```

## Command to reproduce

```
octopus tenant variables list --tenant "Bobs Wood Shop"
```

```
octopus tenant variables view --tenant "Bobs Wood Shop" --name "DatabaseName"
```

## Outcome

```
unknown flag: --tenant
```

## Versions

**cli**: Latest
```
brew info octopus-cli
==> octopusdeploy/taps/octopus-cli: stable 1.7.1
The New CLI (octopus) for Octopus Deploy, a user-friendly DevOps tool for developers that supports release management, deployment automation, and operations runbooks
https://github.com/OctopusDeploy/cli
/usr/local/Cellar/octopus-cli/1.7.1 (6 files, 13.5MB) *
  Built from source on 2023-11-29 at 10:18:30
From: https://github.com/octopusdeploy/homebrew-taps/blob/HEAD/octopus-cli.rb
License: Apache-2.0
```

**Octopus Server**:  https://samples.octopus.app/

ibraheemoctopus:
I've tried to replicate this locally on my end, and I see that when using `octopus tenant variables list tenant "example" I get the expected results; however I try with `octopus tenant variables list --tenant` I see unknown flag: --tenant.

PR #299, opened 2023-11-30
Fix tenant flags for tenant variables list and tenant view

thampton:
Partial fix for https://github.com/OctopusDeploy/cli/issues/297

Adds the --tenant flag as shown in the documentation for 
```
octopus tenant variables list --tenant "example tenant"
```

Adds the --tenant flag as shown in the documentation for
```
octopus tenant view --tenant "example tenant"
```

### DIFF
--- a/pkg/cmd/tenant/view/view.go
+++ b/pkg/cmd/tenant/view/view.go
@@ -1,47 +1,40 @@
 package view
 
 import (
+	"errors"
 	"fmt"
+
 	"github.com/OctopusDeploy/cli/pkg/apiclient"
-	"io"
 
 	"github.com/MakeNowJust/heredoc/v2"
 	"github.com/OctopusDeploy/cli/pkg/constants"
 	"github.com/OctopusDeploy/cli/pkg/factory"
 	"github.com/OctopusDeploy/cli/pkg/output"
-	"github.com/OctopusDeploy/cli/pkg/usage"
 	"github.com/OctopusDeploy/cli/pkg/util/flag"
-	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/client"
 	"github.com/pkg/browser"
 	"github.com/spf13/cobra"
 )
 
 const (
-	FlagWeb = "web"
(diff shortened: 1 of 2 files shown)


pkg/cmd/tenant/view/view.go
```diff
@@ -1,47 +1,40 @@
...
 }
 
-type ViewOptions struct {
-	Client   *client.Client
-	Host     string
-	out      io.Writer
-	idOrName string
-	flags    *ViewFlags
-}
-
```
thampton, line 40: @domenicsim1 I removed this struct because it seemed simplier without it. Should this be retained for unit test purposes?

NickJosevski:
@thampton — an overdue reply, and an apology.

You reported #297 on 29 November 2023 and opened this PR with a fix the very next day. It has sat here without a single comment since. That is a poor way to treat someone who took the trouble to both report a bug and fix it, and we're sorry.

Reviewing it now, the approach was right on both counts: `--tenant` on `tenant variables list` and on `tenant view`, matching what `tenant variables update` already accepted and what the help text advertised at the time.

The branch has unfortunately gone stale in the intervening two and a half years. `tenant variables list` was rewritten around `CommonVariableScopingFeatureToggle` in #548, so the `listRun` this PR modifies no longer exists in that shape and the branch now shows as conflicting. Rather than ask you to rebase work you submitted in 2023, we've carried it forward:

- **#629** — `--tenant` on `tenant view`, your scope, with your approach and a `Co-Authored-By` trailer.
- **#626** — `--tenant` on `tenant variables list`, rebuilt against the rewritten file.

Both add a selector prompt when no tenant is given, and keep the positional argument working. One detail we changed from this PR: rather than dropping `Args: usage.ExactArgs(1)` from `tenant view` outright, we added a `usage.MaximumNArgs(1)` helper, so the positional becomes optional without surplus arguments being silently ignored.

Closing this in favour of those two. Thank you for the report, the fix, and the patience — the delay was ours, not a reflection of the contribution.

PR #300, opened 2023-12-06
chore(docs): remove usage of 'apt-key' from installation instructions

justsomescripts:
This change updates the installation instructions for Debian (based) systems. The usage of `apt-key` for adding keys is deprecated and won't be available in future Debian and Ubuntu versions: https://manpages.debian.org/testing/apt/apt-key.8.en.html (reason: https://unix.stackexchange.com/questions/332672/how-to-add-a-third-party-repo-and-key-in-debian/582853#582853).


justsomescripts:
Fixed in #310

Issue #312, opened 2024-01-16
`project view`, `project-group view` commands don't respect `--output-format` switch


kierenj:
## The bug

The `--output-format` or `-f` option has no effect for `project view` or `project-group view`, despite CLI usage info suggesting otherwise.

## Command to reproduce

```
octopus project view MyProject -f json
octopus project-group view MyProjectGroup -f json
```

## Outcome

```
MyProject (myproject)
Version control branch: Not version controlled
No description provided
View this project in Octopus Deploy: https://XXX.octopus.app//app#/Spaces-1/projects/Projects-XXX


MyProject (MyProjectGroup)
No description provided
  
Projects:
XXX
XXX
XXX

View this project group in Octopus Deploy: https://XXX.octopus.app//app#/Spaces-1/projects?projectGroupId=ProjectGroups-XXX
```

## Versions

**cli**: 2.1.0

**Octopus Server**: How do I find out? It's hosted on octopus.app

brendonthiede:
I see the same issue with `octopus space view "space name" --output-format json`
This is on CLI version 1.7.1 (installed via `go install github.com/OctopusDeploy/cli/cmd/octopus@latest`)
Server version v2024.1 (Build 12401)

Clare-Octopus:
Another user seeing the same issue (internal) - https://octopus.zendesk.com/agent/tickets/223245

PR #318, opened 2024-01-28
Fixes format for project-group view

nate-christensen:
When running octopus project-group view "group" -f json, it wouldn't print in json or anything else for that matter.

This fixes an issue I ran into as well as half of issue #312. I can try to fix the project view one as well if this is merged.

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     services:
       sqlserver:
-        image: mcr.microsoft.com/mssql/server:2019-latest
+        image: mcr.microsoft.com/mssql/server:2019-CU27-ubuntu-20.04
         env:
           ACCEPT_EULA: Y
           SA_PASSWORD: ${{ env.SA_PASSWORD }}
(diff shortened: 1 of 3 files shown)


pkg/cmd/projectgroup/view/view.go
```diff
@@ -73,37 +78,32 @@ func NewCmdView(f factory.Factory) *cobra.Command {
...
-
-	// footer
-	fmt.Fprintf(opts.out, "\nView this project group in Octopus Deploy: %s\n", output.Blue(url))
-
-	if opts.flags.Web.Value {
-		browser.OpenURL(url)
-	}
-
-	return nil
+	return output.PrintArray(allProjects, cmd, output.Mappers[*projects.Project]{
```
tleed5, line 92: For the basic output format, it should return what it did previously, otherwise we are losing some functionality. 

Example of the previous basic output.
```
Default Project Group (ProjectGroups-1)
No description provided

Projects:
NonVcs (nonvcs)
VCS (vcs)

View this project group in Octopus Deploy: http://localhost:8066/app#/Spaces-1/projects?projectGroupId=ProjectGroups-1
```

tleed5 (changes requested):
Looks good overall, just need to add back the output for the basic format 😄

PR #330, opened 2024-02-15
Allow CLI invocation if empty or null arguments

MudraR:
### Why is this change added?
This update lets [octopus](https://octopus.com/docs/octopus-rest-api/cli/octopus) be passed as a command to call `.\octopus.exe`, allowing for empty or null arguments to default to the base `octopus` command

### Before PR
- v2.1.1 installed and added to path:
  image

- going directly to the executable
  image

### After PR
- branch compiled and added to path:
  image

- going directly to the executable
  image

### DIFF
--- a/cmd/octopus/main.go
+++ b/cmd/octopus/main.go
@@ -87,5 +87,5 @@ func main() {
 
 func commandDoesNotRequireClient(args []string) bool {
 	cmdToRun := args[0]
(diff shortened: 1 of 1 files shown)


MudraR:
Bug fixed in https://github.com/OctopusDeploy/cli/pull/336.

PR #333, opened 2024-02-19
chore: Fix small error and warnings in tests, use dockerhub for octopus server image rather than packages.octopushq

borland:
I found that the tests didn't build or run locally because "slices" in `fakesurvey.go` didn't exist; the package name was wrong or had changed.

I fixed that, and a few other small lint warnings while I had it open. I verified all tests pass including integration tests.

I also changed the docker image the tests are using over to the official one on dockerhub rather than our internal docker.packages.octopushq.com -- dockerhub is faster, and we only need the internal one if we are targeting some feature that has not been released to GA self-host customers yet; which is not currently the case.

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -31,7 +31,7 @@ jobs:
           --health-retries 10
           --health-start-period 10s
       octopusserver:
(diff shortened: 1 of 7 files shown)


rain-on (commented):
LGTM - but not sure I'm the one to approve

domenicsim1 (approved):

Issue #110, opened 2022-10-11
Add commands for Runbook operations, specifically 'create-runbook'


dgard1981:
Good morning,

Can I please request that you add additional commands for Runbook operations to the Octopus CLI.

Presently the only Runbook operation in is `run-runbook`. What I'd like to see are these two operations -
- **`create-runbook`:** Ideally this would work in almost exactly the same way as `create-release`, allowing the specification of `defaultPackageVersion` and `package` to set package versions to use with steps. I'd also like to be able to optionally publish the new Runbook after creation.
- **`publish-runbook`:** This operation would allow publishing of an existing Runbook.

Assuming I have one or more steps in both my Deployment Process and a Runbook that references a script, currently when using AWS CodeBuild I can create a package, upload it to S3 and then create a release on Octopus, but I have to manually create the Runbook and then publish it. What I'd like to be able to do is something like this.

```yaml
version: 0.2
env:
  shell: bash
phases:
  ...
  build:
    commands:
      - dotnet-octo pack --id="${PACKAGE_ID}" --basePath="${PACKAGE_BASE_PATH}" --version=${VERSION} --outFolder=Artifacts --format=zip
      - aws s3 cp Artifacts/${PACKAGE_ID}.*.zip ${BUCKET_FEED_URL}
      - dotnet-octo create-release --project ${PROJECT_ID} --space ${SPACE_ID} --defaultPackageVersion ${VERSION} --version ${VERSION} --ignoreExisting
      - dotnet-octo create-runbook --project ${PROJECT_ID} --space ${SPACE_ID} --defaultPackageVersion ${VERSION} --name ${VERSION} --ignoreExisting --publish
```

I'd be happy to provide further details if that would help.

Thanks.
David

toby-freemarket:
This is related to 
https://github.com/OctopusDeploy/cli/issues/165

PR #343, opened 2024-03-15
feat: runbook snapshot create

benPearce1:
This is WIP, not complete

Notes to future me:
This is currently only supporting creating a new snapshot and publishing it - which is a bit wrong.
Publishing an existing snapshot requires POST to the runbook with an update to the `PublishedRunbookSnapshotId` property set to the snapshot id

This should be changed to be two commands:
`runbook snapshot create` with an optional `--publish` parameter
`runbook snapshot publish` to publish an existing snapshot:
  `runbook snapshot publish --runbook "runbook name" --snapshot "Snapshot QWE123"`

will probably fix https://github.com/OctopusDeploy/cli/issues/110

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.21
 require (
 	github.com/AlecAivazis/survey/v2 v2.3.6
 	github.com/MakeNowJust/heredoc/v2 v2.0.1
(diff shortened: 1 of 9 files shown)


benPearce1:
replaced by #514

PR #348, opened 2024-05-09
Order interactive variables by name when deploying releases

EddyMoulton:
When deploying a release with required variables, we ask the user for values via the command line.

Previously the order of these variables was not deterministic, causing the related test to be flakey.
I've now sorted these variable controls alphabetically, which will fix the flakey test and provide a more consistent UX.

[sc-76938]


rain-on (commented):
love it.

rain-on (approved):

PR #355, opened 2024-06-19
fix: Add ability to escape special character when parse variables argument

HuyPhanNguyen:
Support escape `\:`  when parse in `release deploy --variable "Test\:Prompt:Value1"`

BEGIN_COMMIT_OVERRIDE
fix: Add ability to escape special character when parsing variables argument
END_COMMIT_OVERRIDE

### DIFF
--- a/pkg/cmd/runbook/run/run.go
+++ b/pkg/cmd/runbook/run/run.go
@@ -181,7 +181,7 @@ func runbookRun(cmd *cobra.Command, f factory.Factory, flags *RunFlags) error {
 		return err
 	}
 
-	parsedVariables, err := executionscommon.ParseVariableStringArray(flags.Variables.Value, false)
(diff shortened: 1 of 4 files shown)


pkg/cmd/runbook/run/run.go
```diff
@@ -181,7 +181,7 @@ func runbookRun(cmd *cobra.Command, f factory.Factory, flags *RunFlags) error {
 		return err
 	}
 
-	parsedVariables, err := executionscommon.ParseVariableStringArray(flags.Variables.Value)
+	parsedVariables, err := executionscommon.ParseVariableStringArray(flags.Variables.Value, false)
```
benPearce1, line 184: if we just remove the `enableEscape` parameter, this will work for runbook runs too

benPearce1 (changes requested):
Tested this locally, it works but the change needs to be on runbook runs too.

benPearce1 (approved):

PR #362, opened 2024-07-02
Handle `/releases` endpoint default page size of 30

IsaacCalligeros95:
The /releases endpoint has a default page size of 30, we've recently gone over this limit and are dropping older versions in our releases.json. This change bumps the GH action up to 100 and includes the api response (all releases). It's worth considering if we want to support these older releases and whether increasing the releases.json size is worth it.

These older versions have very low download counts (sub 5 in most cases) except 0.5.0 which we download as a part of our GHA installer tests.


benPearce1 (approved):
This seems like a really simple solution. If we have over 100 releases they should start dropping out of support regardless.

PR #364, opened 2024-07-10
feat: add upload build info command

hnrkndrssn:
This PR adds a command to upload build information to Octopus.

```
octopus build-information upload --package-id ThePackage --version 1.2.3 --file /path/to/octopus/buildinformation/file.json --overwrite-mode ignore
```

```
octopus build-info upload ThePackage --version 1.2.3 --file /path/to/octopus/buildinfo/file.json --overwrite-mode fail
```

[sc-83928]

### DIFF
--- a/pkg/cmd/buildinformation/upload/upload.go
+++ b/pkg/cmd/buildinformation/upload/upload.go
@@ -196,20 +196,21 @@ func PromptMissing(opts *UploadOptions) error {
 		}
 	}
 
-	if opts.OverwriteMode.Value == "" {
-		if err := opts.Ask(&survey.Select{
-			Message: "Overwrite mode",
-			Help:    "Determines behavior if the package already exists in the repository. Valid values are 'fail', 'overwrite' and 'ignore'. Default is 'fail'.",
-			Options: []string{
-				"fail",
-				"ignore",
-				"overwrite",
-			},
-			Default: "fail",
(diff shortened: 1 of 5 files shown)


pkg/cmd/buildinformation/upload/upload.go
```diff
@@ -0,0 +1,216 @@
...
+	// 		Options: []string{
+	// 			"fail",
+	// 			"ignore",
+	// 			"overwrite",
+	// 		},
+	// 		Default: "fail",
+	// 	}, &opts.OverwriteMode.Value); err != nil {
+	// 		return err
+	// 	}
+	// }
```
benPearce1, line 213: Does this need to be removed?

hnrkndrssn: yes, it should.

benPearce1 (approved):
LGTM.
It looks like the overwrite mode option is missing from the prompts, probably not a problem as boolean options are hard to reason about when deciding to prompt.

hnrkndrssn:
> LGTM. It looks like the overwrite mode option is missing from the prompts, probably not a problem as boolean options are hard to reason about when deciding to prompt.

As we default it to `fail` if an option isn't specified I removed the prompt and you'd have to add it using `--overwrite-mode` to the command to override the default option.

PR #377, opened 2024-07-12
feat: add create channel command

hnrkndrssn:
```
$ octopus channel create
? Name Testing
? Select the project in which the channel will be created Test project
? Description
? Select the lifecycle to use for the channel Default Lifecycle
? Set channel as default No

Successfully created channel 'Testing', (Channels-5).
View this channel on Octopus Deploy: https://henrik-buildinfo-testing.testoctopus.app/app#/Spaces-1/projects/Projects-1/deployments/channels/edit/Channels-5
octopus channel create --name 'Testing' --project 'Test project' --lifecycle 'Default Lifecycle' --no-prompt
```

[sc-83931]

### DIFF
--- a/pkg/cmd/channel/create/create.go
+++ b/pkg/cmd/channel/create/create.go
@@ -96,15 +96,18 @@ func createRun(opts *CreateOptions) error {
 		return err
 	}
 
-	lifecycle, err := opts.Client.Lifecycles.GetByIDOrName(opts.Lifecycle.Value)
-	if err != nil {
-		return err
-	}
-
 	channel := channels.NewChannel(opts.Name.Value, project.GetID())
 	channel.Description = opts.Description.Value
 	channel.IsDefault = opts.Default.Value
-	channel.LifecycleID = lifecycle.GetID()
+
+	if opts.Lifecycle.Value != "" {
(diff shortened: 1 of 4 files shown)


pkg/cmd/channel/create/create.go
```diff
@@ -0,0 +1,187 @@
...
+	opts.Project.Value = selectedProject.Name
+
+	err = question.AskDescription(opts.Ask, "", "channel", &opts.Description.Value)
+	if err != nil {
+		return err
+	}
+
+	var selectedLifecycle *lifecycles.Lifecycle
+	if opts.Lifecycle.Value == "" {
+		selectedLifecycle, err = selectors.Lifecycle("Select the lifecycle to use for the channel", opts.Client, opts.Ask)
```
benPearce1, line 157: I think this need be modified to support the `Inherit from project` option

hnrkndrssn: Yes, you are right. I forgot about that option.

benPearce1 (changes requested):
It doesn't prompt for the project if I don't supply a project identifier

benPearce1 (approved):

PR #383, opened 2024-07-30
fix: support slug for executions API

hnrkndrssn:


hnrkndrssn:
Not required as we already do a lookup by id/slug/name as part of the release create/deploy commands.

PR #388, opened 2024-08-12
feat: list all projects in any one group.

nate-christensen:
Allows one to list all projects in a specific group

### DIFF
--- a/pkg/cmd/projectgroup/project-group.go
+++ b/pkg/cmd/projectgroup/project-group.go
@@ -5,6 +5,7 @@ import (
 	createCmd "github.com/OctopusDeploy/cli/pkg/cmd/projectgroup/create"
 	deleteCmd "github.com/OctopusDeploy/cli/pkg/cmd/projectgroup/delete"
 	listCmd "github.com/OctopusDeploy/cli/pkg/cmd/projectgroup/list"
+	cmdListProjects "github.com/OctopusDeploy/cli/pkg/cmd/projectgroup/projects"
 	viewCmd "github.com/OctopusDeploy/cli/pkg/cmd/projectgroup/view"
 	"github.com/OctopusDeploy/cli/pkg/constants"
 	"github.com/OctopusDeploy/cli/pkg/constants/annotations"
@@ -30,6 +31,7 @@ func NewCmdProjectGroup(f factory.Factory) *cobra.Command {
 	cmd.AddCommand(listCmd.NewCmdList(f))
 	cmd.AddCommand(deleteCmd.NewCmdList(f))
 	cmd.AddCommand(viewCmd.NewCmdView(f))
+	cmd.AddCommand(cmdListProjects.NewCmdListProjects(f))
 
 	return cmd
 }
--- a/pkg/cmd/projectgroup/projects/list/list.go
+++ b/pkg/cmd/projectgroup/projects/list/list.go
@@ -1,0 +1,102 @@
+package list
+
+import (
+	"github.com/MakeNowJust/heredoc/v2"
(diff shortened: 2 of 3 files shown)


NickJosevski:
Thanks for this, and apologies it has sat unreviewed for so long.

I've had a proper look, and rather than ask you to iterate on this shape I want to explain where we landed, because I think the underlying capability is worth having — just somewhere else.

### Where we'd rather put it

`octopus project-group view` already lists the projects in a group, both in `basic` and in `json`:

```
$ octopus project-group view "Default Project Group" -f basic
Default Project Group (ProjectGroups-1)
Projects:
Calamari parity probe (calamari-parity-probe)
cli-tests (cli-tests)
...

$ octopus project-group view "Default Project Group" -f json | jq '.Projects[].Name'
```

So the scripting case is largely covered today, and a new `project-group projects list` command would mostly duplicate it.

What *is* genuinely missing is discoverability: someone who wants "the projects in group X" reaches for `octopus project list`, not for `project-group`. So we're adding the filter there instead:

```
octopus project list --group 'Default Project Group'
octopus project list -g ProjectGroups-1
```

That matches `project create -g/--group` and `project clone -g/--group`, avoids a three-level `project-group projects list` whose middle noun would exist to hold a single verb, and needs the same single API call your version makes (the projects collection has no `projectGroupId` filter, so it goes via the group's `Projects` link either way — which also gives us pagination for free).

I'm putting that up as a separate change and will link it here. To be clear, the idea came from this PR, so thank you for it.

### Two things worth knowing regardless

Whichever shape wins, these bit us when we ran your branch, and they're useful feedback:

1. **The flag registration has its arguments swapped.** In `pkg/cmd/projectgroup/projects/list/list.go`:

   ```go
   flags.StringVarP(&listFlags.Group.Value, "group", "g", "filter packages to match only ones that contain the given string", "")
   ```

   The signature is `(p, name, shorthand, value, usage)`, so that string (copy-pasted from `package list`) becomes the flag's *default value* and the help text ends up empty:

   ```
   FLAGS
    -g, --group string    (default "filter packages to match only ones that contain the given string")
   ```

   Running it with no `--group` then looks up a group by that sentence and fails with `cannot find the item`.

2. **No prompt when the identifier is missing.** Our list commands that hang off a parent resource take the identifier positionally *as well as* by flag, and fall back to a selector when it isn't supplied, erroring clearly only under `--no-prompt` — see `pkg/cmd/project/branch/list/list.go`. We recently made that consistent across the CLI in #627.

Also worth noting for anyone reading later: it does build and vet cleanly against current `main`, and the happy path works correctly in table, json and basic output.

My suggestion is that we land `project list --group` and close this one out, but I'll leave it open for now in case you'd like to weigh in first. Either way, I'd genuinely welcome more contributions — this was a real gap you spotted.

NickJosevski:
Closing this out now, in favour of #689, which adds the same capability as a `--group`/`-g` filter on `octopus project list`:

```
octopus project list --group 'Default Project Group'
octopus project list -g ProjectGroups-1
```

The reasoning is in my comment above — short version: it puts the filter where people already look for projects, matches the `-g/--group` flag `project create` and `project clone` use, and makes the same single API call via the group's `Projects` link that your branch did.

To be clear about the provenance: the gap was yours to spot, and #689 credits you in its description. Leaving it a few weeks in case you wanted to weigh in first; closing now so the queue reflects reality rather than because the input wasn't wanted.

The two notes above — the swapped `StringVarP` arguments, and the missing positional/prompt fallback — are worth carrying into any future contribution here, since they'd apply to any command in this shape.

Thanks again, and please do send more our way.

PR #392, opened 2024-08-29
feat: add ability to change git refs when creating release

APErebus:
Updates the SDK to `v2.51.1` which adds support for changing the git refs when creating a release.

Adds a new flag `--git-resource` which can be used multiple times to specify the git resource refs as part of the `release create` command.

The string format is similar to the packages, but a bit simpler.

`{StepName}:{GitRef}` - This targets the primary git resource for a step only
`{StepName}:{GitResourceName}:{GitRef}` - This targets a secondary git resource

Wildcards (`*`) can be used for the `{GitRef}`. In this case, the step-defined default branch will be used.

This also includes the normal interactive flow where the user can specify what git resource refs they want

Sample interactive output

```bash
? Select the project in which the release will be created Git refs
Selecting only available channel 'Default'.
STEP NAME               GIT RESOURCE  GIT REF
Deploy Kubernetes YAML  <primary>     main
? Git resource ref override string (y to accept, u to undo, r to reset, ? for help): Deploy Kubernetes YAML:refs/head/test

? Git resource ref override string (y to accept, u to undo, r to reset, ? for help): Deploy Kubernetes YAML:refs/head/test
STEP NAME               GIT RESOURCE  GIT REF
Deploy Kubernetes YAML  <primary>     refs/head/test
? Git resource ref override string (y to accept, u to undo, r to reset, ? for help): u

? Git resource ref override string (y to accept, u to undo, r to reset, ? for help): u
STEP NAME               GIT RESOURCE  GIT REF
Deploy Kubernetes YAML  <primary>     main
? Git resource ref override string (y to accept, u to undo, r to reset, ? for help): Deploy Kubernetes YAML:refs/tags/1.0.0
                                                                                                           
? Git resource ref override string (y to accept, u to undo, r to reset, ? for help): Deploy Kubernetes YAML:refs/tags/1.0.0
STEP NAME               GIT RESOURCE  GIT REF
Deploy Kubernetes YAML  <primary>     refs/tags/1.0.0
? Git resource ref override string (y to accept, u to undo, r to reset, ? for help): y

? Git resource ref override string (y to accept, u to undo, r to reset, ? for help): y
```

Shortcut story: [sc-69100]


APErebus:
This is not ready for review or merging as I missed all the logic around asking the user if they want to override git refs (like pacakges)

pkg/cmd/release/create/gitResources.go
```diff
@@ -0,0 +1,301 @@
...
+
+	for _, gitResource := range deploymentProcessTemplate.GitResources {
+		result = append(result, &GitResourceGitRef{
+			ActionName:      gitResource.ActionName,
+			GitRef:          gitResource.DefaultBranch,
+			GitResourceName: gitResource.Name,
+		})
+	}
+
+	return result, nil
```
APErebus, line 55: Don't need to return an error

pkg/cmd/release/create/gitResources.go
```diff
@@ -0,0 +1,301 @@
...
+	return result
+}
+
+type GitResourceGitRef struct {
+	ActionName      string
+	GitRef          string
+	GitResourceName string
+}
+
+const GitResourceOverrideQuestion = "Git resource ref override string (y to accept, u to undo, r to reset, ? for help):"
```
MissedTheMark, line 64:
```suggestion
const GitResourceOverrideQuestion = "Git resource reference override string (y to accept, u to undo, r to reset, ? for help):"
```
I know that `git ref` is a valid and common term to use, but I think this would be easier to read/understand if we used the full word `reference` instead.  Same goes in the help text.  If we were just saying `git ref` it'd be fine, but adding our `resource` to it makes it a bit clunky.

Open to disagreement though, this is very subjective 😄

pkg/cmd/release/create/gitResources.go
```diff
@@ -0,0 +1,301 @@
...
+			}
+		}
+		// loop around and let them put in more input
+	}
+	return overriddenGitResourceGitRefs, nil
+}
+
+func ParseGitResourceGitRefString(gitResourceRef string) (*GitResourceGitRef, error) {
+	if strings.TrimSpace(gitResourceRef) == "" {
+		return nil, errors.New("empty git resource git ref specification")
```
MissedTheMark, line 164: We just call them `git resource ref` in the help output.  TBH I'm not really sure which we should use.

pkg/cmd/release/create/gitResources.go
MissedTheMark: Or maybe we should use `Git resource git ref` like you do in `ParseGitResourceGitRefString`?

MissedTheMark: `Artifact git ref` would make it much better 😜

pkg/cmd/release/create/gitResources.go
```diff
@@ -0,0 +1,301 @@
...
+
+	actionName = strings.TrimSpace(components[0])
+	switch len(components) {
+	case 2:
+		gitRef = strings.TrimSpace(components[1])
+	case 3:
+		gitResourceName = strings.TrimSpace(components[1])
+		gitRef = strings.TrimSpace(components[2])
+	default:
+		return nil, fmt.Errorf("git resource git ref specification \"%s\" does not use expected format", gitResourceRef)
```
MissedTheMark, line 182:
```suggestion
		return nil, fmt.Errorf("git resource git ref specification \"%s\" does not use an expected format", gitResourceRef)
```
Maybe?

pkg/cmd/release/create/gitResources.go
```diff
@@ -0,0 +1,301 @@
...
+
+func ApplyGitResourceOverrides(gitResourcesBaseline []*GitResourceGitRef, overrides []*GitResourceGitRef) []*GitResourceGitRef {
+	result := make([]*GitResourceGitRef, 0, len(gitResourcesBaseline))
+
+	for _, gitResourceGitRef := range gitResourcesBaseline {
+
+		//Try and find any override
+		overrideIdx := slices.IndexFunc(overrides, func(override *GitResourceGitRef) bool {
+			return override.Equals(gitResourceGitRef)
+		})
```
MissedTheMark, line 232: I may have misunderstood how the overrides work, but aren't they a stack? So if we find the first match here, could we be missing more recent overrides for the same ref?

There doesn't seem to be a test for this scenario that I can see either.

pkg/cmd/release/create/gitResources_test.go
```diff
@@ -0,0 +1,414 @@
...
+		{input: "Action:Name1:*", expect: &create.GitResourceGitRef{ActionName: "Action", GitRef: "*", GitResourceName: "Name1"}},
+		{input: "Action=Name1=refs/heads/main", expect: &create.GitResourceGitRef{ActionName: "Action", GitRef: "refs/heads/main", GitResourceName: "Name1"}},
+		{input: "Action=Name1=*", expect: &create.GitResourceGitRef{ActionName: "Action", GitRef: "*", GitResourceName: "Name1"}},
+
+		//Invalid inputs
+		{input: "", isErrorExpected: true},
+		{input: "    ", isErrorExpected: true},
+		{input: "Action", isErrorExpected: true},
+		{input: ":refs/heads/main", isErrorExpected: true},
+		{input: "::refs/heads/main", isErrorExpected: true},
```
MissedTheMark, line 34: What happens if you mix the delimiters? eg `Action=Name1:main`?

MissedTheMark (approved):
Good work, there's a lot in this! 
Just a couple of formatting comments, and one question around how the override stack works when applying.

pkg/cmd/release/create/gitResources.go
```diff
@@ -0,0 +1,301 @@
...
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/deployments"
+	"golang.org/x/exp/slices"
+	"io"
+	"strings"
+)
+
+var gitResourceGitRefLoopHelpText = heredoc.Doc(`
+bold(GIT RESOURCE SELECTION)
+ This screen presents the list of Git resources used by your project, and the steps
+ which reference them.
```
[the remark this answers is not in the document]
APErebus: Just to match the package formatting TBH

pkg/cmd/release/create/gitResources.go
```diff
@@ -0,0 +1,301 @@
...
+ This screen presents the list of Git resources used by your project, and the steps
+ which reference them. 
+
+bold(COMMANDS)
+ Any any point, you can enter one of the following:
+ - green(?) to access this help screen
+ - green(y) to accept the list of Git resources and proceed with creating the release
+ - green(u) to undo the last edit you made to Git resource Git refs
+ - green(r) to reset all Git resource Git ref edits
+ - A Git resource Git ref override string.
```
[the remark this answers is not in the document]
APErebus: Fixed in 79b2a54

pkg/cmd/release/create/gitResources.go
APErebus: I standardized on `git resource git reference`

pkg/cmd/release/create/gitResources_test.go
APErebus: I _think_ it actually still works. I'll add a test and check

APErebus: It does work, but it's not consistent with the server-side code (which only supports using one delimiter type at once).
Updated in adc6c26 to be consistent. Also learned what `rune` represents (`char`)

pkg/cmd/release/create/gitResources.go
APErebus: oo, good point. Let me check that

APErebus: Fixed in 3f90de7

pkg/cmd/release/create/gitResources_test.go
MissedTheMark: IKR, it's so weird

PR #391, opened 2024-08-19
feat: support --version flag

benPearce1:
fixes #269 
image

image

other commands still work
image

PR #398, opened 2024-09-16
fix: no program args now outputs the root help again

APErebus:
`2.8.0` included #391, however, if you just do `octopus` with no args, it no longer prints the root help output.

`2.7.0`
{0D39E626-5E39-4E0F-805A-09EE0B7C5E86}

`2.8.0`
{03207CB0-76FD-44BF-A37F-34E0F75A7605}

Fixed
{B54B986A-5B88-423F-B396-F29BA784A602}

This PR fixes that by returning `Help` from the root command if the version parameter is not set (previously it was returning `nil`)


MissedTheMark (approved):
✅

PR #401, opened 2024-09-16
Update homebrew deployment process

APErebus:
Uses latest worker tools and also uses correct authentication origin


MissedTheMark (approved):
LTGM

PR #408, opened 2024-09-18
Update deployment process to publish updates to winget

APErebus:
Uses the [winget-create](https://github.com/microsoft/winget-create) tool to create a new update of the `OctopusDeploy.Cli` manifests.

I also updated the RPM & APT steps to pull the `publish-linux` via Artifactory to avoid Docker rate-limiting issues (as Artifactory will cache the image). This is not a long term plan, but just alleviates the current issue.

I can't really test this, so will wait for the next release of the CLI to see the results


MissedTheMark (approved):
LTGM, 🤞 it works! #yolo

PR #409, opened 2024-09-18
Removes a whitespace between Path key and value when creating a zip p…

mik-ky:
…ackage with json output format. Additionally prints a new line so it's consistent with the other output options.

This is to make it consistent with our other json output formats, like release creation or deploy.

image

### DIFF
--- a/pkg/cmd/package/nuget/create/create.go
+++ b/pkg/cmd/package/nuget/create/create.go
@@ -151,14 +151,15 @@ func createRun(cmd *cobra.Command, opts *NuPkgCreateOptions) error {
 	}
 
 	nuget, err := pack.BuildPackage(opts.PackageCreateOptions, outFilePath)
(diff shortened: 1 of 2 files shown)


pkg/cmd/package/nuget/create/create.go
```diff
@@ -151,14 +151,15 @@ func createRun(cmd *cobra.Command, opts *NuPkgCreateOptions) error {
 	}
 
 	nuget, err := pack.BuildPackage(opts.PackageCreateOptions, outFilePath)
-	if (nuget != nil) {
+	if nuget != nil {
 		switch outputFormat {
-			case constants.OutputFormatBasic:
```
mik-ky, line 156: These are some auto-formatting changes from my IDE.

susanpann (approved):
🙅🏾 no spaces allowed here

PR #427, opened 2024-10-01
Do not prompt for space name (for client) on space creation

IsaacCalligeros95:
Do not prompt for existing space when creating a new space.
[sc-73470]
image

### DIFF
--- a/pkg/cmd/space/create/create.go
+++ b/pkg/cmd/space/create/create.go
@@ -48,7 +48,7 @@ type CreateOptions struct {
 }
 
 func NewCreateOptions(f factory.Factory, flags *CreateFlags, c *cobra.Command) *CreateOptions {
-	dependencies := cmd.NewDependencies(f, c)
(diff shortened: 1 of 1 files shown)


pkg/cmd/space/create/create.go
```diff
@@ -48,7 +48,7 @@ type CreateOptions struct {
 }
 
 func NewCreateOptions(f factory.Factory, flags *CreateFlags, c *cobra.Command) *CreateOptions {
-	dependencies := cmd.NewDependencies(f, c)
+	dependencies := cmd.NewSystemDependencies(f, c)
```
benPearce1, line 51: Was that all it was!! Noice!

benPearce1 (approved):

pkg/cmd/space/create/create.go
IsaacCalligeros95: Yep 😆, I'll leave this to be bundled in the next release someone needs to push out.

PR #440, opened 2024-11-01
chore: Separate package version and Git resource functions from create release

geofflamrock:
This PR is part of a series that adds support to the CLI for interacting with runbooks stored in Git for version controlled projects

- https://github.com/OctopusDeploy/cli/pull/440
- https://github.com/OctopusDeploy/cli/pull/441
- https://github.com/OctopusDeploy/cli/pull/443
- https://github.com/OctopusDeploy/cli/pull/444

## Background 

Git runbooks operate differently to runbooks stored in the Octopus database in that there are no draft/published snapshots, these are replaced with running a runbook using a Git reference such as `main`. This means that in order to run a runbook that uses packages or Git resources these need to be supplied as part of the run creation.

## Changes

This PR moves some of the common functions for resolving package versions and Git resources from the `releases/create` package into their own packages. 

## Reducing risk

There are lots of tests for creating releases which have had only mechanical changes, these should all pass and give us confidence that this change is good. We could look to separate some of the tests for package version resolution in this PR, however I've opted to tackle that separately to give us confidence in this initial change.


pkg/packages/packages.go
```diff
@@ -0,0 +1,613 @@
...
+
+	// used to locate the deployment process VersioningStrategy Donor Package
+	PackageReferenceName string
+}
+
+// BuildPackageVersionBaseline loads the deployment process template from the server, and for each step+package therein,
+// finds the latest available version satisfying the channel version rules. Result is the list of step+package+versions
+// to use as a baseline. The package version override process takes this as an input and layers on top of it
+func BuildPackageVersionBaseline(octopus *octopusApiClient.Client, packages []releases.ReleaseTemplatePackage,
+	setAdditionalFeedQueryParameters func(releases.ReleaseTemplatePackage, feeds.SearchPackageVersionsQuery) (feeds.SearchPackageVersionsQuery, error)) ([]*StepPackageVersion, error) {
```
geofflamrock, line 69: This `func` parameter has a change in it to make it generic and usable when creating Git runbooks and creating releases. When creating a release we need to check the channel rules for the associated package and add any additional search parameters to ensure that the version selected satisfies the rules. This parameter allows us to do that.

pkg/packages/packages.go
```diff
@@ -0,0 +1,613 @@
...
+	// these 3 fields are the main ones for showing the user
+	PackageID  string
+	ActionName string // "StepName is an obsolete alias for ActionName, they always contain the same value"
+	Version    string // note this may be an empty string, indicating that no version could be found for this package yet
+
+	// used to locate the deployment process VersioningStrategy Donor Package
+	PackageReferenceName string
+}
+
+// BuildPackageVersionBaseline loads the deployment process template from the server, and for each step+package therein,
```
geofflamrock, line 65: Oops this comment isn't correct

PR #448, opened 2024-11-20
fix: disable/enable tenant command docs

grace-rehn:
Generating docs for CLI resulted in incorrect docs for `tenant enable` and `tenant disable` - https://github.com/OctopusDeploy/docs/pull/2565

Fix this by updating `Use` and `Example` to include  the name/id argument.


IsaacCalligeros95 (approved):
LGTM

Issue #479, opened 2025-02-06
Running a Runbook will prompt for a variable value for a variable that is not scoped to that Runbook


dan-close:
## The bug

The `runbook run` command will prompt the user for a value if there is a prompted variable in the project that is scoped to a different Runbook.

## Command to reproduce

```
octopus runbook run --project 'Test Project' --name 'Test Runbook 2' --environment 'Development'
```

## Outcome

```
? Runbook1.Test.Variable
```

## Versions

**cli**: 2.14.2

**Octopus Server**: 2024.4.7132

PR #480, opened 2025-02-07
fix: Check runbook variable context when prompt variable

HuyPhanNguyen:
Fix Variable Scope Checking in Runbook Run Command

Previously, when executing the `runbook run` command, the CLI would prompt users for all variables marked as "Prompt", regardless of their scope. This meant users were being asked to provide values for variables that weren't actually relevant to their current context.

For example, if a variable was scoped to Runbook2 or Process of ProjectA, but the user was running the runbook1, they would still be prompted for that variable's value.
## Result
Fixes [sc-103259](https://app.shortcut.com/octopusdeploy/story/103259/sev-3-running-a-runbook-with-the-cli-will-prompt-user-for-a-variable-value-if-there-is-a-prompted-variable-scoped-to-a-different)
Fixes [#479](https://github.com/OctopusDeploy/cli/issues/479)
 
 
## Before
image

## After
image

### DIFF
--- a/pkg/cmd/runbook/run/run.go
+++ b/pkg/cmd/runbook/run/run.go
@@ -598,7 +598,7 @@ func AskDbRunbookRunQuestions(octopus *octopusApiClient.Client, stdout io.Writer
 		}
 	}
 
-	options.Variables, err = askRunbookPreviewVariables(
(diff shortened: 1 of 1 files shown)


benPearce1 (approved):
Looks good. Thanks

benPearce1 (approved):

PR #488, opened 2025-02-17
List scoped common variables

bec-callow-oct:
[SC-93474]

When scoped tenant common variables are enabled, both common and project variables can be scoped by multiple environments. This PR provides support for variables with an array of environment scopes.

## Before
Tenant common variables had no environment scoping, so Environment was blank. Project variables were only scoped to one environment, so per template, one row existed per environment.

```
$ ./octopus tenant variables list "t1"
NAME                        LABEL         TYPE     OWNER                    ENVIRONMENT  VALUE                   SENSITIVE  DEFAULT VALUE
template1                                 Library  set1                                  default                 false      true
secret template                           Library  set1                                  ***                     true       true
missing t1                                Library  missing set 1                         <missing>                      false      true
missing t2                                Library  missing set 1                         default                 false      true
proj template1                            Project  var scopes               Dev          default                 false      true
proj template1                            Project  var scopes               Test         CLI missing update 1    false      false
proj template1                            Project  var scopes               Prod         default                 false      true
Missing Proj template 1     missing vars  Project  var scopes               Test         <missing>               false      true
Missing Proj template 1     missing vars  Project  var scopes               Prod         <missing>               false      true
Missing Proj template 1     missing vars  Project  var scopes               Dev          <missing>               false      true
```

## After
Tenant common variables now have environment scoping, and both common and project variables can have multiple environment scopes. We still list one row per template and environment because each environment can only have one value. Tested combining environment names for shared scopes into a comma separated list, but values were cut off when the list was long.

```
$ ./octopus tenant variables list "t1"
NAME                        LABEL  TYPE     OWNER                    ENVIRONMENT  VALUE                   SENSITIVE  DEFAULT VALUE
missing t1                         Library  missing set 1            Prod         <missing>               false      true
missing t1                         Library  missing set 1            Test         <missing>               false      true
missing t1                         Library  missing set 1            Dev          Dev scope only          false      false
missing t2                         Library  missing set 1            Dev          default                 false      true
missing t2                         Library  missing set 1            Test         default                 false      true
missing t2                         Library  missing set 1            Prod         default                 false      true
no scope common template 1         Library  no scope set1            Prod         default                 false      true
no scope common template 1         Library  no scope set1            Dev          default                 false      true
no scope common template 1         Library  no scope set1            Test         default                 false      true
secret template                    Library  set1                     Prod         ***                     true       true
secret template                    Library  set1                     Test         ***                     true       true
secret template                    Library  set1                     Dev          ***                     true       true
template1                          Library  set1                     Test         default                 false      true
template1                          Library  set1                     Prod         default                 false      true
template1                          Library  set1                     Dev          latest cli              false      false
Missing Proj template 1            Project  var scopes               Prod         <missing>               false      true
Missing Proj template 1            Project  var scopes               Dev          <missing>               false      true
Missing Proj template 1            Project  var scopes               Test         <missing>               false      true
proj template1                     Project  var scopes               Test         update 1                false      false
proj template1                     Project  var scopes               Dev          default                 false      true
proj template1                     Project  var scopes               Prod         default                 false      true
proj template 1                    Project  var scopes 2             Dev          project-template1-test  false      false
proj template 1                    Project  var scopes 2             Test         project-template1-test  false      false
```

### DIFF
--- a/pkg/cmd/tenant/variables/list/list.go
+++ b/pkg/cmd/tenant/variables/list/list.go
@@ -323,17 +323,26 @@ func unwrapProjectVariablesV1(variables variables.ProjectVariable, environmentMa
 
 func unwrapCommonVariables(variable variables.TenantCommonVariable, environmentMap map[string]string) []*VariableValue {
 	var results []*VariableValue = nil
+	var isDefault = variable.ID == ""
+	var displayValue string
+
+	if isDefault {
+		displayValue = getDisplayValue(variable.Template.DefaultValue)
+	} else {
+		displayValue = getDisplayValue(&variable.Value)
+	}
+
 	for _, environmentId := range variable.Scope.EnvironmentIds {
 
 		results = append(results, &VariableValue{
 			Type:            LibraryVariableSetType,
 			OwnerName:       variable.LibraryVariableSetName,
 			Name:            variable.Template.Name,
-			Value:           getDisplayValue(&variable.Value),
+			Value:           displayValue,
 			IsSensitive:     variable.Value.IsSensitive,
-			IsDefaultValue:  variable.ID == "", // Variables without an ID are sourced from the MissingVariables list. For consistency with V1, IsDefault applies to both default and missing variables
+			IsDefaultValue:  isDefault, // Variables without an ID are sourced from the MissingVariables list. For consistency with V1, IsDefault applies to both default and missing variables
 			ScopeName:       environmentMap[environmentId],
-			HasMissingValue: variable.Value.Value == "",
+			HasMissingValue: isDefault && displayValue == "",
 		})
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/cmd/tenant/variables/list/list.go
```diff
@@ -99,67 +102,169 @@ func listRun(cmd *cobra.Command, f factory.Factory, id string) error {
...
+	toggleValue, err := featuretoggle.IsToggleEnabled(client, "CommonVariableScopingFeatureToggle")
 
-	missingVariablesResponse, err := client.Tenants.GetMissingVariables(variables.MissingVariablesQuery{TenantID: vars.TenantID})
-	var missingVariables []variables.MissingVariable
-	if len(*missingVariablesResponse) > 0 {
-		missingVariables = (*missingVariablesResponse)[0].MissingVariables
-	}
-	var allVariableValues []*VariableValue
-	for _, element := range vars.LibraryVariables {
+	if toggleValue {
```
tothegills, line 107: I was going to recommend splitting this into two functions, one for toggle on and off. But perhaps this makes more sense if we are going to remove on half anyway.

pkg/util/featuretoggle/feature_toggles.go
```diff
@@ -0,0 +1,21 @@
...
+// IsToggleEnabled retrieves an Octopus feature toggle by name. If an error occurs, it returns false.
+func IsToggleEnabled(client *client.Client, toggleName string) (bool, error) {
+	toggleRequest := &configuration.FeatureToggleConfigurationQuery{
+		Name: toggleName,
+	}
+
+	returnToggleResponse, err := configuration.Get(client, toggleRequest)
+
+	if err != nil {
+		return false, err
```
tothegills, line 17: What happens when the toggle name parameter does not exist? ie the cli is being run against a version of Octopus that doesn't have the toggle filter implemented.

tothegills (approved):
I've tested with feature toggles on and off. The correct variables are shown in both cases.
I think the choice to print a row per environment is a good one given our constraints. We can iterate on the design if feedback indicates that it is confusing.

I have a suspicion that we may end up with the wrong answer when asking if the feature toggle is disabled for versions of Octopus that do not have the feature toggle filter implemented. We may need to get a name match (and handle when the name does not exist).

pkg/util/featuretoggle/feature_toggles.go
bec-callow-oct: Good catch. I've updated the function to handle an empty list of toggles, as well as a name check since older versions of octopus return all toggles instead of just one

PR #496, opened 2025-03-04
Nj/6.4 fix

NickJosevski:
$Env:DOTNET_ROOT = "C:\Users\ContainerAdministrator\AppData\Local\Microsoft\dotnet"

is a fix until we can work out what's special about either wingetcreate, this version of botnet and how it's being installed in a our worker-tools container image.


MissedTheMark (approved):
Looks reasonable as a fix 👍

PR #522, opened 2025-05-23
Delete winget-pkgs fork before running winget-create

APErebus:
We've discovered if the fork repo is too out of sync, the `winget-create update` call can fail.

This change deletes the fork repo before `update` call runs, which will re-fork the repo as part of it's execution


wlthomson (approved):
Thanks for this 👍 .

PR #584, opened 2026-03-26
fix: Support escaped delimiters in --package for Maven packages

### LINKED ISSUE #547: Maven Package Names with Colons Break --package Flag Parsing
## The bug
The `--package` flag cannot handle Maven package names that contain colons (e.g., `com.example:my-package`) when creating releases. The current parsing logic splits on all colons, causing Maven packages like `com.example:my-package:1.0.0` to be incorrectly parsed as having 3 components (`com.example`, `my-package`, `1.0.0`) instead of the expected 2 components (`com.example:my-package`, `1.0.0`).
This breaks the package override functionality for Maven repositories and other package types that use colons in their package identifiers.
## Command to reproduce
```bash
octopus release create --project "MyProject" --package "com.example:my-package:1.0.0"
```
## Outcome
The command fails to parse the package override string correctly, treating `com.example:my-package:1.0.0` as a 3-component string instead of a 2-component `package:version` format. This results in an error during package override parsing.

**Expected behavior**: The Maven package `com.example:my-package` should be recognized as a single package identifier, with `1.0.0` as the version.

**Actual behavior**: The colon in the Maven package name is treated as a delimiter, breaking the parsing logic and causing the command to fail or incorrectly resolve package versions.

## Potential Root Cause
The issue appears to be in the `splitPackageOverrideString` function in `pkg/packages/packages.go` (lines 223-225):
```go
func splitPackageOverrideString(s string) []string {
    return util.SplitString(s, []int32{':', '/', '='})
}
```
This function splits on all colons without considering that package names themselves may contain colons, which is common in Maven's `groupId:artifactId` naming convention.
The package override format expects:
- 2 components: `{package}:{version}` or `{step}:{version}`
- 3 components: `{step}:{package}:{version}` or `{package-ref}:{package}:{version}`
But Maven packages like `com.example:my-package:1.0.0` get split into 3 components when they should be treated as 2.

## Versions
**cli**: 2.18.0

HuyPhanNguyen:
## Summary
- Add backslash escape support (`\:` `\/` `\=`) to `splitPackageOverrideString` so Maven package IDs containing colons can be correctly specified via `--package`
- Escape delimiter characters in `ToPackageOverrideString` so the reconstructed string sent to the server API is also parseable
- Add test cases for Maven-style package IDs, escaped slashes in step names, and escaped backslashes

Fixes https://github.com/OctopusDeploy/cli/issues/547
Related: FD-135
Related: HPY-1293

**Requires companion server PR:** OctopusDeploy/OctopusDeploy#41734

## Context
Maven package IDs use the format `Group:Artifact` (e.g. `com.juliusbaer.fi-master:deployment`). The CLI's `--package` argument uses `:` as a delimiter between package ID and version, making it impossible to specify Maven packages correctly.

With this fix, users can escape the colon:
```
octopus release create --package "com.juliusbaer.fi-master\:deployment:25.2026.04.1"
```

## Test plan
- [x] Existing `ParsePackageOverrideString` tests pass (no regression)
- [x] Existing `ToPackageOverrideString` tests pass (no regression)
- [x] Existing `ResolvePackageOverride` and `ApplyPackageOverride` tests pass
- [x] New tests for escaped colons in Maven package IDs
- [x] New tests for escaped slashes in step names
- [x] New tests for escaped backslashes
- [ ] Manual test with Maven feed

### DIFF
--- a/pkg/cmd/release/create/create_test.go
+++ b/pkg/cmd/release/create/create_test.go
@@ -2427,9 +2427,8 @@ func TestReleaseCreate_ToPackageOverrideString(t *testing.T) {
 		{name: "maven-pkg-ref-ver", input: &packages.PackageVersionOverride{PackageID: "com.juliusbaer.fi-master:deployment", PackageReferenceName: "ref", Version: "25.2026.04.1"}, expect: `com.juliusbaer.fi-master\:deployment:ref:25.2026.04.1`},
 		// Step name with slash gets escaped
 		{name: "step-slash-ver", input: &packages.PackageVersionOverride{ActionName: "Deploy Templates/templates", Version: "1.0"}, expect: `Deploy Templates\/templates:1.0`},
-		// Literal backslashes in package IDs/step names get escaped as \\
-		{name: "pkg-with-backslash", input: &packages.PackageVersionOverride{PackageID: `foo\bar`, Version: "1.0"}, expect: `foo\\bar:1.0`},
-		{name: "step-with-trailing-backslash", input: &packages.PackageVersionOverride{ActionName: `step\`, Version: "1.0"}, expect: `step\\:1.0`},
+		// Literal backslashes pass through unescaped (wire compatibility with pre-FD-135 servers)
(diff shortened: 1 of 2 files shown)


hnrkndrssn (approved):
LGTM 👍

PR #591, opened 2026-04-17
Fix public doc linting issues

flin-8:
Fixes [MD-1770: Fix broken CLI docs update workflow linting](https://linear.app/octopus/issue/MD-1770/fix-broken-cli-docs-update-workflow-linting)

The generated docs PR: https://github.com/OctopusDeploy/docs/pull/3109/changes

### DIFF
--- a/cmd/gen-docs/main.go
+++ b/cmd/gen-docs/main.go
@@ -201,7 +201,7 @@ navOrder: {{.Position}}
 import SamplesInstance from 'src/shared-content/samples/samples-instance.include.md';
 
 {{.Command.Long}}
-` + "\n```" + `{{define "T1"}}Usage:{{if .Runnable}}
+` + "\n```bash" + `{{define "T1"}}Usage:{{if .Runnable}}
   {{.UseLine}}{{end}}{{if .HasAvailableSubCommands}}
   {{.CommandPath}} [command]{{end}}{{if gt (len .Aliases) 0}}
 
@@ -233,7 +233,7 @@ Use "{{.CommandPath}} [command] --help" for more information about a command.{{e
 ## Examples
 
 <SamplesInstance />
-` + "\n```" + `
+` + "\n```bash" + `
 {{ .Command.Example }}
 ` + "\n```\n" + `
 {{- end }}
--- a/cmd/gen-docs/main.go
+++ b/cmd/gen-docs/main.go
@@ -241,7 +241,8 @@ Use "{{.CommandPath}} [command] --help" for more information about a command.{{e
 ## Learn more
 
 - [Octopus CLI](/docs/octopus-rest-api/cli)
-- [Creating API keys](/docs/octopus-rest-api/how-to-create-an-api-key)`
+- [Creating API keys](/docs/octopus-rest-api/how-to-create-an-api-key)
+`
 
 const indexTemplate = `---
 layout: src/layouts/Default.astro
@@ -263,7 +264,8 @@ The Octopus CLI is built and maintained by the Octopus Deploy team, but it is al
 
 ## Commands {#octopusCommandLine-Commands}
 
-` + "\n`octopus` supports the following commands:\n" +
+` + "`octopus` supports the following commands:" +
 	`
 {{range .Pages}}
-- **[{{.Title}}]({{.RelativePath}})**:  {{.Command.Short}}.{{end}}`
+- **[{{.Title}}]({{.RelativePath}})**:  {{.Command.Short}}.{{end}}
+`
--- a/cmd/gen-docs/main.go
+++ b/cmd/gen-docs/main.go
@@ -201,7 +201,7 @@ navOrder: {{.Position}}
 import SamplesInstance from 'src/shared-content/samples/samples-instance.include.md';
 
 {{.Command.Long}}
-` + "\n```bash" + `{{define "T1"}}Usage:{{if .Runnable}}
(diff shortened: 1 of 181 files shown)


cmd/gen-docs/main.go
```diff
@@ -201,7 +201,7 @@ navOrder: {{.Position}}
 import SamplesInstance from 'src/shared-content/samples/samples-instance.include.md';
 
 {{.Command.Long}}
-` + "\n```" + `{{define "T1"}}Usage:{{if .Runnable}}
+` + "\n```text" + `{{define "T1"}}Usage:{{if .Runnable}}
```
flin-8, line 204: [MD040](https://github.com/DavidAnson/markdownlint/blob/main/doc/md040.md) fenced-code-language - Fenced code blocks should have a language specified

cmd/gen-docs/main.go
```diff
@@ -233,15 +233,16 @@ Use "{{.CommandPath}} [command] --help" for more information about a command.{{e
...
 {{ .Command.Example }}
 ` + "\n```\n" + `
 {{- end }}
 
 ## Learn more
 
 - [Octopus CLI](/docs/octopus-rest-api/cli)
-- [Creating API keys](/docs/octopus-rest-api/how-to-create-an-api-key)`
+- [Creating API keys](/docs/octopus-rest-api/how-to-create-an-api-key)
+`
```
flin-8, line 245: [MD047](https://github.com/DavidAnson/markdownlint/blob/main/doc/md047.md) single-trailing-newline - Files should end with a single newline character

cmd/gen-docs/main.go
```diff
@@ -263,7 +264,8 @@ The Octopus CLI is built and maintained by the Octopus Deploy team, but it is al
 
 ## Commands {#octopusCommandLine-Commands}
 
-` + "\n`octopus` supports the following commands:\n" +
+` + "`octopus` supports the following commands:" +
```
flin-8, line 267: [MD012](https://github.com/DavidAnson/markdownlint/blob/main/doc/md012.md) no-multiple-blanks - Multiple consecutive blank lines

pkg/cmd/account/aws/create/create.go
```diff
@@ -67,7 +68,7 @@ func NewCmdCreate(f factory.Factory) *cobra.Command {
 		Use:     "create",
 		Short:   "Create an AWS account",
 		Long:    "Create an AWS account in Octopus Deploy",
-		Example: heredoc.Docf("$ %s account aws create", constants.ExecutableName),
+		Example: heredoc.Docf("%s account aws create", constants.ExecutableName),
```
flin-8, line 71: [MD014 - Dollar signs used before commands without showing output](https://github.com/DavidAnson/markdownlint/blob/main/doc/md014.md)

pkg/cmd/project/variables/variables.go
```diff
@@ -22,9 +22,9 @@ func NewCmdVariables(f factory.Factory) *cobra.Command {
 		Short:   "Manage project variables",
 		Long:    "Manage project variables in Octopus Deploy",
 		Example: heredoc.Docf(`
-			$ %[1]s project variable list "Deploy Web App"
-			$ %[1]s project variable view --name "DatabaseName" --project Deploy
-			$ %[1]s project variable update
+			%[1]s project variable list "Deploy Web App"
+			%[1]s project variable view --name "DatabaseName" --project Deploy
+			%[1]s project variable update
```
flin-8, line 27: [MD014](https://github.com/DavidAnson/markdownlint/blob/main/doc/md014.md) commands-show-output - Dollar signs used before commands without showing output

zentron (approved):
Scanning through the changes it all _looks_ reasonable, im assuming the format must have changed at some point since it was last generated. 🤷

pkg/cmd/project/variables/variables.go
flin-8: @zentron if we think it's important to keep the `$` at the front, then we can override the linting rule

NickJosevski:
Picking up on the suggestion to automate the docs update — worth noting the workflow already runs on `v*` tag pushes (see `.github/workflows/generate-docs.yml`), so docs do regenerate on every release. The gap is between releases, where they can drift.

**Recommendation:** add push-to-main with a path filter, plus an empty-diff guard so we don't open no-op PRs. Preferable to a `schedule:` because it catches doc-affecting changes promptly and avoids running CI when nothing relevant changed.

Trigger:

```yaml
on:
  workflow_dispatch:
  push:
    branches:
      - main
    tags:
      - 'v*'
    paths:
      - 'cmd/gen-docs/**'
      - 'pkg/cmd/**'
      - 'go.mod'
      - 'go.sum'
      - '.github/workflows/generate-docs.yml'
```

Then a `Check for docs changes` step after `Generate cli docs` that runs `git status --porcelain` in the docs checkout, sets a `changed` output, and gates the Commit / Push / Create PR steps on `if: steps.docs-diff.outputs.changed == 'true'`.

A couple of notes:
- GitHub does not apply `paths` filters to tag pushes, so the existing release behaviour is preserved.
- The branch name uses a timestamp (`enh-cliupdate-octopus-$(date ...)`), so frequent merges could pile up open PRs in the docs repo. If that becomes noisy a follow-up could reuse a rolling branch and update the existing PR.

A `schedule:` (e.g. weekly) is a fine fallback if we'd rather decouple from commit cadence, but it'll still need the same empty-diff guard and will lag merges by up to a week.

PR #594, opened 2026-04-29
Use package version for Winget publish

sathvikkumar-octo:
Don't use `Octopus.Release.Number` for the Winget push as it means if there is an error deploying the release we can't easily create a new release to try again and need to delete the old release first.

This brings it inline with the other external repositories we are deploying the CLI to


APErebus (approved):
Good idea!

PR #605, opened 2026-05-22
ci: regenerate docs on push to main with empty-diff guard

NickJosevski:
## Goal

Run the docs-update workflow on every push to `main` so docs don't drift between releases. The change is just the trigger: adding `branches: [main]` and a `paths:` filter scoped to dirs that affect generated docs (`cmd/gen-docs/**`, `pkg/cmd/**`, `go.mod`, `go.sum`, the workflow file). Net diff is 9 lines.

Follow-up from #591.

## Why specify `branches: [main]` rather than omitting the filter?

GitHub's `push:` filter is asymmetric — specifying `tags:` alone *implicitly excludes* all branch pushes. The previous config listed only `tags: ['v*']`, so no branch push ever triggered the workflow. `paths:` filters don't apply to tag pushes, so the existing `v*` release trigger continues to fire unconditionally. ([GitHub docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetbranchesbranches-ignore): *"If you only define one of the two (tags or branches), the workflow won't run for events affecting the undefined Git ref."*)

## Test plan

- [ ] Merge a change under `pkg/cmd/**` and confirm a docs PR is opened against `OctopusDeploy/docs`.
- [ ] Push a `v*` tag (or wait for the next release) and confirm the release-time docs PR still opens.
- [ ] Confirm a push to `main` that only touches unrelated paths (e.g. `README.md`) does not trigger the workflow.

### DIFF
--- a/.github/workflows/generate-docs.yml
+++ b/.github/workflows/generate-docs.yml
@@ -2,8 +2,16 @@ name: 'Generate Docs'
 on:
   workflow_dispatch:
   push:
+    branches:
+      - main
     tags:
       - 'v*'
+    paths:
+      - 'cmd/gen-docs/**'
+      - 'pkg/cmd/**'
+      - 'go.mod'
+      - 'go.sum'
+      - '.github/workflows/generate-docs.yml'
 env:
   DOCS_TOKEN: ${{ secrets.DOCS_TOKEN }}
 jobs:


benPearce1 (approved):

NickJosevski:
https://linear.app/octopus/issue/MD-1772/automate-go-cli-doc-updates

PR #606, opened 2026-05-22
Preserve --variable values through interactive runbook run

NickJosevski:
Closes #582.

## The bug

`octopus runbook run -v Name:Value` silently drops the variables in interactive mode — both from the printed `Automation Command` and from the actual run, which falls back to default values. `--no-prompt` was the workaround.

Repro from #582:

```
octopus runbook run -p "New Project" -n "Runbook" -s "Kenny" -v "certCommonName:CLI" -v "prompt:CLI" -e "Development"
```

The `Automation Command` comes back missing both `-v` values:

```
Automation Command: octopus runbook run --project 'New Project' --name 'Runbook' --environment 'Development' --no-prompt
```

## cause

__The result map was rebuilt from the controls, not from the caller's inputs.__ Inside `askRunbookPreviewVariables` (`pkg/cmd/runbook/run/run.go`), the `result` map was built fresh by iterating the form preview's `flattenedControls`, only adding entries for controls. Any CLI-supplied variable without a matching control was discarded. The caller then overwrote `options.Variables` with this stripped map.

`--no-prompt` skips this codepath entirely — that's why the workaround works.

## Fix

Extract the variable-resolution into a pure helper `resolveRunbookPreviewVariables`, and seed the result map with the caller's CLI variables _before_ processing controls. Variables matching a control are still canonicalised to the control's name (existing case-fixing behaviour preserved); variables without a matching control pass through unchanged.

## Tests

New `pkg/cmd/runbook/run/preview_variables_test.go` (internal `package run`) with 6 unit tests: passthrough (the fix), casing canonicalisation, no-prompt-when-supplied, sensitive tracking, prompt-fires-for-missing-required, and a combination case.

## Tests

- [x] CI passes `go test ./pkg/cmd/runbook/run/...`
- [x] Manual repro from #582 . `-v` values are in both the `Automation Command` and the actual run.

### DIFF
--- a/pkg/cmd/runbook/run/preview_variables_test.go
+++ b/pkg/cmd/runbook/run/preview_variables_test.go
@@ -1,0 +1,108 @@
+package run
+
+import (
+	"testing"
+
+	"github.com/AlecAivazis/survey/v2"
+	"github.com/AlecAivazis/survey/v2/core"
+	"github.com/OctopusDeploy/cli/pkg/question"
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/deployments"
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/resources"
+	"github.com/stretchr/testify/assert"
+)
+
+func noPromptAsker(t *testing.T) question.Asker {
+	t.Helper()
+	return func(p survey.Prompt, response interface{}, opts ...survey.AskOpt) error {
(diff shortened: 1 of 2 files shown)


NickJosevski:
@benPearce1 if/when you have time because you have (some) context in this area, does this fix look ok to you?

NickJosevski:
superseded by https://github.com/OctopusDeploy/cli/pull/665

PR #609, opened 2026-06-02
feat: add channel list, view, and delete commands

NickJosevski:
Adds three subcommands to `octopus channel`:

- **`list`** — list a project's channels. Takes the project positionally or via `-p/--project`, with `--filter`/`-q` for name filtering.
- **`view`** — view a channel by name or ID. Resolves the lifecycle name and shows version rules, Git reference/resource rules, tenant tags, custom field definitions and ephemeral-environment settings. Supports `--web`.
- **`delete`** — delete a channel by name or ID. Confirmation prompt in interactive mode, `-y/--confirm` for automation.

All three support `table`/`json`/`basic` output and prompt for the project when it isn't supplied.

## Notes

- **Config-as-Code**: `delete` warns in interactive mode when the project is version-controlled — the server does not guard against deleting a channel an OCL deployment references.
- **Lookup scoping**: `view` and `delete` resolve the channel through a shared helper that falls back to a project-scoped lookup, so an ID belonging to another project is never used.
- **No slug lookup**: the server does not resolve channel slugs on the by-ID route, and `go-octopusdeploy` v2.108.0 doesn't expose `Slug` on `channels.Channel`. Both commands accept name or ID only; surfacing slugs needs an SDK bump.
- **`view` vs `create`**: `view` is read-only and shows fields `channel create` cannot yet set. Extending `create` (and adding `update`) is a follow-up, not part of this PR.

tested against local instance

### DIFF
--- a/pkg/cmd/channel/delete/delete.go
+++ b/pkg/cmd/channel/delete/delete.go
@@ -1,0 +1,175 @@
+package delete
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/AlecAivazis/survey/v2"
+	"github.com/MakeNowJust/heredoc/v2"
+	"github.com/OctopusDeploy/cli/pkg/apiclient"
+	"github.com/OctopusDeploy/cli/pkg/constants"
+	"github.com/OctopusDeploy/cli/pkg/factory"
+	"github.com/OctopusDeploy/cli/pkg/output"
+	"github.com/OctopusDeploy/cli/pkg/question"
+	"github.com/OctopusDeploy/cli/pkg/question/selectors"
+	"github.com/OctopusDeploy/cli/pkg/util/flag"
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/channels"
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/client"
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/projects"
+	"github.com/spf13/cobra"
+)
+
+const (
+	FlagProject = "project"
+)
+
+type DeleteFlags struct {
+	Project *flag.Flag[string]
+	*question.ConfirmFlags
+}
+
+func NewDeleteFlags() *DeleteFlags {
+	return &DeleteFlags{
+		Project:      flag.New[string](FlagProject, false),
+		ConfirmFlags: question.NewConfirmFlags(),
+	}
+}
+
+type DeleteOptions struct {
+	Client   *client.Client
+	Ask      question.Asker
+	Out      *cobra.Command
+	NoPrompt bool
+	IdOrName string
+	*DeleteFlags
+}
+
+func NewCmdDelete(f factory.Factory) *cobra.Command {
+	deleteFlags := NewDeleteFlags()
+	cmd := &cobra.Command{
+		Use:     "delete {<name> | <id> | <slug>}",
+		Short:   "Delete a channel",
+		Long:    "Delete a channel in Octopus Deploy",
+		Aliases: []string{"del", "rm", "remove"},
+		Example: heredoc.Docf(`
+			%[1]s channel delete "Hotfix" --project myProject
+			%[1]s channel rm Channels-123 --project myProject -y
+		`, constants.ExecutableName),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if deleteFlags.Project.Value == "" {
+				return errors.New("--project is required")
+			}
+
+			c, err := f.GetSpacedClient(apiclient.NewRequester(cmd))
+			if err != nil {
+				return err
+			}
+
+			idOrName := ""
+			if len(args) > 0 {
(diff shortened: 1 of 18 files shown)


pkg/cmd/channel/delete/delete.go
```diff
@@ -0,0 +1,162 @@
...
+
+	flags := cmd.Flags()
+	flags.StringVarP(&deleteFlags.Project.Value, deleteFlags.Project.Name, "p", "", "Name or ID of the project the channel belongs to")
+	question.RegisterConfirmDeletionFlag(cmd, &deleteFlags.Confirm.Value, "channel")
+
+	return cmd
+}
+
+func deleteRun(opts *DeleteOptions) error {
+	project, err := selectors.ResolveProject(opts.Client, opts.Ask, !opts.NoPrompt,
```
YuKitsune, line 88: The changes to the other `delete` commands here have a comment saying `in automation mode we validate the flags up front, before making any API calls`. Let's keep the behavior consistent.

pkg/cmd/channel/shared/shared.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/client"
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/projects"
+)
+
+// ResolveChannel finds a channel belonging to project, by ID or name. A direct
+// GetByID is tried first; if the caller supplied a name, or the ID belongs to a
+// channel in a different project, we fall back to the project-scoped lookup so we
+// never return a channel from a project the caller didn't ask for.
+func ResolveChannel(octopus *client.Client, project *projects.Project, idOrName string) (*channels.Channel, error) {
+	if channel, err := octopus.Channels.GetByID(idOrName); err == nil && channel != nil && channel.ProjectID == project.GetID() {
```
YuKitsune, line 34: The other "Resolve" methods include the `promptEnabled` check, why not this one too?

pkg/cmd/channel/delete/delete.go
```diff
@@ -0,0 +1,162 @@
...
+	return &DeleteFlags{
+		Project:      flag.New[string](FlagProject, false),
+		ConfirmFlags: question.NewConfirmFlags(),
+	}
+}
+
+type DeleteOptions struct {
+	Client   *client.Client
+	Ask      question.Asker
+	Out      *cobra.Command
```
YuKitsune: This shouldn't be called `Out`. I'd suggest using `*cmd.Dependencies` here like in the other options structs.

pkg/cmd/channel/delete/delete.go
```diff
@@ -0,0 +1,162 @@
...
+	if err != nil {
+		return err
+	}
+
+	// In interactive mode, warn for version-controlled (CaC) projects since deleting a
+	// channel can break OCL deployments referencing it.
+	if !opts.NoPrompt && project.IsVersionControlled {
+		opts.Out.Printf("%s This project is version-controlled (Config-as-Code). Deleting this channel can break OCL deployments that reference it.\n",
+			output.Yellow("Warning:"))
+	}
```
YuKitsune: Is this warning necessary? We don't have this for any other version controlled things.

pkg/cmd/channel/list/list.go
```diff
@@ -0,0 +1,135 @@
...
+				listFlags.Project.Value = args[0]
+			}
+
+			return listRun(cmd, f, listFlags)
+		},
+	}
+
+	flags := cmd.Flags()
+	flags.StringVarP(&listFlags.Project.Value, listFlags.Project.Name, "p", "", "Name or ID of the project to list channels for")
+	flags.StringVarP(&listFlags.Filter.Value, listFlags.Filter.Name, "q", "", "filter channels to match only ones with a name containing the given string")
```
YuKitsune:
```suggestion
	flags.StringVarP(&listFlags.Filter.Value, listFlags.Filter.Name, "q", "", "Filter channels to match only ones with a name containing the given string")
```

pkg/cmd/channel/list/list.go
```diff
@@ -0,0 +1,135 @@
...
+	viewModels := make([]ChannelViewModel, 0, len(allChannels))
+	for _, c := range allChannels {
+		if filter != "" && !strings.Contains(strings.ToLower(c.Name), filter) {
+			continue
+		}
+		viewModels = append(viewModels, ChannelViewModel{
+			ID:          c.ID,
+			Name:        c.Name,
+			Description: c.Description,
+			LifecycleID: c.LifecycleID,
```
YuKitsune, line 112: In the `view` command, `Inherited from project` is shown when `LifecycleId == ""`, should we do something similar here?

YuKitsune (changes requested):

NickJosevski:
@YuKitsune I actioned what you brought up

YuKitsune (approved):

PR #612, opened 2026-07-07
Fix: allow completion command to run without an Octopus login

### LINKED ISSUE #611: octopus completion powershell returns root help/banner instead of PowerShell completion script (exit code 3)
## The bug

The PowerShell autocomplete generation command appears to be broken in Octopus CLI 2.21.3 and 2.21.2.

The CLI advertises a completion command:

```text
completion: Generate the autocompletion script for the specified shell
```

The following help command succeeds and shows the expected usage:

```powershell
octopus help completion powershell
```

Output:

```text
Generate the autocompletion script for powershell.

To load completions in your current shell session:

        octopus completion powershell | Out-String | Invoke-Expression
```

However, running the documented command:

```powershell
octopus completion powershell
```

does not generate a PowerShell completion script.

Instead, it outputs the top-level Octopus CLI banner/help beginning with:

```text
Work seamlessly with Octopus Deploy from the command line.
```

and exits with code:

```text
3
```

No output is written to STDERR.

Because the command emits the root help text rather than a PowerShell script, the documented installation command also fails:

```powershell
octopus completion powershell | Out-String | Invoke-Expression
```

Result:

```text
Invoke-Expression : The term 'Work' is not recognized as the name of a cmdlet,
function, script file, or operable program.
```

## Command to reproduce

```powershell
octopus --version

octopus help completion powershell

octopus completion powershell

$LASTEXITCODE
```

Also:

```powershell
octopus completion powershell | Out-String | Invoke-Expression

$LASTEXITCODE
```

## Outcome

Actual:

```text
Work seamlessly with Octopus Deploy from the command line.

                ####
            ############
         ##################
         ...
```

Exit code:

```text
3
```

STDERR:

```text
<empty>
```

Expected:

```text
A PowerShell completion script is written to STDOUT.
```

## Versions

**cli**:

```text
2.21.3
```

Also reproduced on:

```text
2.21.2
```

**Octopus Server**:

```text
Not applicable
```

The completion script generation command should not require an Octopus Server connection to function.

## Links

None.

HuyPhanNguyen:
# Background

`octopus completion <shell>` is a local Cobra command that prints a shell completion script; it never contacts the server. But `commandDoesNotRequireClient` in `cmd/octopus/main.go` doesn't include it, so when the user hasn't run `octopus login` the client factory fails to build and the CLI exits with code 3, printing the config banner to stdout instead of the script. Piping that into `Invoke-Expression` then fails with `The term 'Work' is not recognized`. The same gap hits Cobra's hidden `__complete`/`__completeNoDesc` helpers, which the shell calls on every Tab press — so completions stay broken when not logged in even after installing them.

# Results

Adds `completion`, `__complete`, and `__completeNoDesc` to `commandDoesNotRequireClient` so completion works without a login.

Fixes FD-555
Fixes https://github.com/OctopusDeploy/cli/issues/611

## Before

Not logged in:

    $ octopus completion powershell
    Work seamlessly with Octopus Deploy from the command line.
    ... (banner) ...
    # exit code 3, nothing on stderr

`octopus completion powershell | Out-String | Invoke-Expression` → `The term 'Work' is not recognized`.

## After

Not logged in:

    $ octopus completion powershell
    # powershell completion for octopus ...
    Register-ArgumentCompleter -CommandName 'octopus' -ScriptBlock $__octopusCompleterBlock
    # exit code 0

`octopus __complete ""` also returns the command list (exit 0) instead of failing.

# Testing

- Added a unit test covering the completion commands in `commandDoesNotRequireClient`.
- Manually verified against a build: unauthenticated `completion powershell` and `__complete` now exit 0; the `... | Out-String | Invoke-Expression` command runs clean.

> Note: this fixes the not-logged-in failure only. A separate report that tab-completion doesn't fire in Windows PowerShell 5.1 (when logged in) is not addressed here.

### DIFF
--- a/cmd/octopus/main.go
+++ b/cmd/octopus/main.go
@@ -96,5 +96,5 @@ func commandDoesNotRequireClient(args []string) bool {
 		return true
 	}
 	cmdToRun := args[0]
-	return cmdToRun == "config" || cmdToRun == "version" || cmdToRun == "--version" || cmdToRun == "-v" || cmdToRun == "help" || cmdToRun == "login" || cmdToRun == "logout" || (cmdToRun == "package" && util.SliceContains(args, "create"))
+	return cmdToRun == "config" || cmdToRun == "version" || cmdToRun == "--version" || cmdToRun == "-v" || cmdToRun == "help" || cmdToRun == "login" || cmdToRun == "logout" || cmdToRun == "completion" || cmdToRun == "__complete" || cmdToRun == "__completeNoDesc" || (cmdToRun == "package" && util.SliceContains(args, "create"))
 }
--- a/cmd/octopus/main_test.go
+++ b/cmd/octopus/main_test.go
@@ -1,0 +1,33 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCommandDoesNotRequireClient_CompletionCommandsDoNotRequireClient(t *testing.T) {
+	cases := [][]string{
+		{"completion"},
+		{"completion", "powershell"},
+		{"completion", "bash"},
+		{"__complete", "completion", "powershell", ""},
+		{"__completeNoDesc", "release", "list", ""},
+	}
+
+	for _, args := range cases {
+		assert.True(t, commandDoesNotRequireClient(args), "expected %v to not require a client", args)
+	}
+}
+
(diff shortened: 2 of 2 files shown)


wlthomson (approved):
Limited blast radius + tests = LGTM.

PR #616, opened 2026-07-16
housekeeping, ci: bump GitHub Actions to Node 24 runtimes

NickJosevski:
Move Node-based actions off the deprecated Node 20 runtime (GitHub is force-mapping them to Node 24) by bumping each to the lowest major version that ships a Node 24 runtime:

### DIFF
--- a/.github/workflows/generate-docs.yml
+++ b/.github/workflows/generate-docs.yml
@@ -23,12 +23,12 @@ jobs:
         run: echo "BRANCH_NAME=enh-cliupdate-octopus-$(date +'%Y%m%d-%H%M%S')" >> $GITHUB_ENV
 
       - name: Checkout CLI
-        uses: actions/checkout@v3
+        uses: actions/checkout@v5
         with:
           path: cli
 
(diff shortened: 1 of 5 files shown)


.github/workflows/generate-docs.yml
```diff
@@ -43,16 +43,16 @@ jobs:
           git checkout -b $BRANCH_NAME
 
       - name: Set up Go
-        uses: actions/setup-go@v3
+        uses: actions/setup-go@v6
         with:
-          go-version: 1.21
+          go-version-file: cli/go.mod
```
APErebus, line 48: What's this change doing?

APErebus (approved):
Lgtm, just 1 question

.github/workflows/generate-docs.yml
NickJosevski: avoids drift and maintaining version number in more places, it's the 2nd line in the `go.mod` file, and this mechanism gets it from there.

CleanShot 2026-07-16 at 13 21 40

PR #623, opened 2026-08-04
fix: emit valid JSON from package create commands

### LINKED ISSUE #518: JSON output containing file paths invalid
## The bug

When using `--output-format json` with `package zip create` command, the json returned contains single backslash characters which are not legal according to json.org spec.

## Command to reproduce
Use the octopus cli to create a package specifying json as the output format:

```
octopus package zip create --id 'test.package' --version '0.0.1' --base-path 'c:\temp\packageme' --out-folder 'c:\temp\' --include '**' --overwrite --no-prompt --output-format json | Out-String
```

## Outcome
The above command returns the file path of the package in json
```
{"Path":"c:\temp\test.package.0.0.1.zip"}
```

Note the single backslash characters (notably the \t which are control characters).  I would expect the output to be escaped, as:
```
{"Path":"c:\\temp\\test.package.0.0.1.zip"}
```

When parsing the unescaped output via powershell, the file path is corrupted as the `\t` in the above path is interpreted as control characters.

To workaround the issue you can do the escaping yourself:
```
$result = $output -replace '\\', '\\\\' | ConvertFrom-Json
```

## Versions

**cli**: 2.16.0

**Octopus Server**: n/a

NickJosevski:
Fixes #518

## Problem

`package zip create` and `package nuget create` built their JSON output with a `Printf` format string:

```go
cmd.Printf(`{"Path":"%s"}`, zip.Name())
```

Any backslash in the path was written through unescaped. On Windows that produces output which is not legal JSON, and which corrupts the path when parsed — `\t` becomes a tab:

```
{"Path":"c:\temp\test.package.0.0.1.zip"}
```

## Change

Marshal with `encoding/json` instead. Both commands now go through a single `pack.PrintPackageCreated` helper — the duplicated inline formatting is precisely what let the same bug exist in two places.

Basic and table output are unchanged.

## Verification

Unit tests assert the JSON round-trips and that basic/table stay unescaped. End-to-end against a directory containing a backslash:

```
$ octopus package zip create ... --output-format json
{"Path":".../packtest/we\\ird/test.package.0.0.1.zip"}

$ python3 -c "import json; print(json.load(open('out.json'))['Path'])"
.../packtest/we\ird/test.package.0.0.1.zip
```

Previously this failed to parse.

### DIFF
--- a/pkg/cmd/package/support/pack_test.go
+++ b/pkg/cmd/package/support/pack_test.go
@@ -33,10 +33,13 @@ func TestPrintPackageCreated_JsonEscapesBackslashes(t *testing.T) {
 
 	// The output must be legal JSON; previously the path was interpolated
 	// directly into a string literal, leaving raw backslashes behind.
+	assert.Equal(t, `{"Path":"c:\\temp\\test.package.0.0.1.zip"}`+"\n", output)
+
+	// And it must parse back to the path we were given, not a mangled one:
+	// unescaped, the \t above would have been read as a tab.
 	var result PackageCreatedOutput
 	assert.NoError(t, json.Unmarshal([]byte(output), &result))
 	assert.Equal(t, windowsPackagePath, result.Path)
-	assert.Contains(t, output, `c:\\temp\\test.package.0.0.1.zip`)
 }
 
 func TestPrintPackageCreated_BasicIsUnescaped(t *testing.T) {
--- a/pkg/cmd/package/support/pack_test.go
+++ b/pkg/cmd/package/support/pack_test.go
@@ -2,9 +2,12 @@ package support
 
 import (
 	"bytes"
+	"encoding/json"
 	"errors"
+	"github.com/OctopusDeploy/cli/pkg/constants"
 	flag "github.com/OctopusDeploy/cli/pkg/util/flag"
 	"github.com/OctopusDeploy/cli/test/testutil"
+	"github.com/spf13/cobra"
 	"github.com/stretchr/testify/assert"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


pkg/cmd/package/support/pack_test.go
```diff
@@ -13,6 +16,41 @@ import (
...
+
+func TestPrintPackageCreated_JsonEscapesBackslashes(t *testing.T) {
+	output := capturePrintPackageCreated(constants.OutputFormatJson, windowsPackagePath)
+
+	// The output must be legal JSON; previously the path was interpolated
+	// directly into a string literal, leaving raw backslashes behind.
+	var result PackageCreatedOutput
+	assert.NoError(t, json.Unmarshal([]byte(output), &result))
+	assert.Equal(t, windowsPackagePath, result.Path)
+	assert.Contains(t, output, `c:\\temp\\test.package.0.0.1.zip`)
```
YuKitsune: It'd be nice to check the entire JSON string here, rather than just the path.

YuKitsune (approved):

pkg/cmd/package/support/pack_test.go
NickJosevski: Done in df1ff20 — now asserts the whole document:

```go
assert.Equal(t, `{"Path":"c:\\temp\\test.package.0.0.1.zip"}`+"\n", output)
```

Good catch: `Contains` would have passed on any output that merely contained the escaped path, so a malformed document around it would have slipped through. Kept the unmarshal round-trip below it, since that is what pins down the actual regression — unescaped, the `\t` parses back as a tab.

I checked the assertion is not vacuous by temporarily restoring the old `Printf` form; it fails with `{"Path":"c:\temp\test..."}` vs the escaped expectation.

PR #632, opened 2026-08-04
fix: produce valid NuGet packages from package nuget create

### LINKED ISSUE #477: `octopus package nuget create` generates invalid NuGet packages
## The bug

The `octopus package nuget create` command generates nupkg files without metadata required by certain NuGet external feed providers supported by Octopus Deploy ([e.g. Cloudsmith](https://octopus.com/docs/packaging-applications/package-repositories/guides/cloudsmith-feed)) to treat them as valid. When pushing these NuGet packages to the external feed, the request is rejected.

The old `dotnet.octopus.cli` tool (`dotnet octo`) created valid NuGet packages that could be pushed to feeds like Cloudsmith. This tool [appears to be deprecated](https://octopus.com/docs/octopus-rest-api/octopus-cli).

There also exists a NuGet CLI tool (`nuget.exe`), however this [no longer works on Ubuntu 24.04](https://github.com/NuGet/setup-nuget/issues/168) as it relies on Mono, which is [no longer supported](https://www.mono-project.com/).

The newer `dotnet pack` tooling does not support creating NuGet packages out of plain files as is expected by Octopus Deploy's feeds. (It relies on a .csproj project being passed in to generate a .NET library package)

## Command to reproduce

```bash
mkdir -p input
touch input/test.txt
octopus package nuget create --id test --version 0.1 --author "Required to generate nuspec" --base-path input --out-folder . --verbose --no-prompt
dotnet nuget push -s "https://<Cloudsmith feed url>/v3/index.json" -k "<Cloudsmith API key>" test.0.1.nupkg
```

## Outcome

Here's the console output that happens right now:

```
Pushing test.0.1.nupkg to 'https://<Cloudsmith feed url>/v2/package'...
  PUT https://<Cloudsmith feed url>/v2/package/
  BadRequest https://<Cloudsmith feed url>/v2/package/ 867ms
error: Response status code does not indicate success: 400 (Bad Request).
```

Using Fiddler, I caught the HTTP response body to be the following:

```json
{"ok": false, "error": "Could not upload artifact: Failed to get .rels XML from NuGet package and find .nuspec XML"}
```

When using the older tool `dotnet octo pack` instead, valid NuGet packages are created and these can be pushed to NuGet external feeds like Cloudsmith.

## Versions

**cli**: 2.14.2

**Octopus Server**: N/A

NickJosevski:
Fixes #477

## Problem

`package nuget create` wrote the nuspec and the matched files into a zip and named it `.nupkg`. A `.nupkg` is an [Open Packaging Conventions](https://learn.microsoft.com/en-us/dotnet/standard/io/) container, not a plain zip, so the result was missing every OPC part:

```
$ unzip -l Acme.Widget.1.2.3.nupkg     # before
  Acme.Widget.nuspec
  lib/
  lib/thing.dll
  readme.txt
```

Opened with .NET's OPC reader — which is what strict feeds use — that package resolves to **zero parts** and has no manifest relationship at all:

```
--- before
    OPC open           : OK (0 parts)
    manifest rel       : MISSING
    core-properties rel: MISSING
    package creator    :
```
## Change

See `BuildPackageWithContents` and `buildArchive`

## Verification

**Structure now matches `dotnet pack` exactly.** 

```
dotnet pack                              octopus package nuget create
  _rels/.rels                              _rels/.rels
  Ref.Widget.nuspec                        Acme.Widget.nuspec
  lib/net8.0/ref.dll                       lib/thing.dll
  [Content_Types].xml                      [Content_Types].xml
  package/services/.../<hash>.psmdcp       package/services/.../<hash>.psmdcp
```

**Validated with a real .NET OPC reader** (`System.IO.Packaging` + `NuGet.Packaging`), before and after:

```
--- before                             --- after
    OPC open           : OK (0 parts)      OPC open           : OK (6 parts)
    manifest rel       : MISSING           manifest rel       : /Acme.Widget.nuspec
    core-properties rel: MISSING           core-properties rel: /package/services/.../<hash>.psmdcp
    package creator    :                   package creator    : Acme
```

### DIFF
--- a/pkg/cmd/package/nuget/create/create.go
+++ b/pkg/cmd/package/nuget/create/create.go
@@ -150,7 +150,14 @@ func createRun(cmd *cobra.Command, opts *NuPkgCreateOptions) error {
 		fmt.Fprintf(opts.Writer, "\nAutomation Command: %s\n", autoCmd)
 	}
 
-	nuget, err := pack.BuildPackage(opts.PackageCreateOptions, outFilePath)
+	// a .nupkg is an OPC container, not a plain zip; it needs its content types,
+	// relationships and core properties or feeds will reject it
+	nuget, err := pack.BuildPackageContents(opts.PackageCreateOptions, outFilePath, &pack.PackageContents{
+		ExcludeDirectories: true,
+		ExtraEntries: func(paths []string) ([]pack.ArchiveEntry, error) {
+			return buildOpcParts(opts.Id.Value, opts.Version.Value, opts.Author.Value, opts.Description.Value, paths)
+		},
+	})
 	if nuget != nil {
 		switch outputFormat {
 		case constants.OutputFormatBasic:
--- a/pkg/cmd/package/nuget/create/create.go
+++ b/pkg/cmd/package/nuget/create/create.go
@@ -152,7 +152,7 @@ func createRun(cmd *cobra.Command, opts *NuPkgCreateOptions) error {
 
 	// a .nupkg is an OPC container, not a plain zip; it needs its content types,
 	// relationships and core properties or feeds will reject it
-	nuget, err := pack.BuildPackageContents(opts.PackageCreateOptions, outFilePath, &pack.PackageContents{
+	nuget, err := pack.BuildPackageWithContents(opts.PackageCreateOptions, outFilePath, &pack.PackageContents{
 		ExcludeDirectories: true,
 		ExtraEntries: func(paths []string) ([]pack.ArchiveEntry, error) {
 			return buildOpcParts(opts.Id.Value, opts.Version.Value, opts.Author.Value, opts.Description.Value, paths)
--- a/pkg/cmd/package/nuget/create/opc.go
+++ b/pkg/cmd/package/nuget/create/opc.go
@@ -1,0 +1,210 @@
+package create
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/xml"
+	"fmt"
+	"path"
+	"sort"
+	"strings"
+
+	pack "github.com/OctopusDeploy/cli/pkg/cmd/package/support"
+	"github.com/OctopusDeploy/cli/pkg/constants"
+)
+
+// A .nupkg is not a plain zip. It is an Open Packaging Conventions container,
+// and feeds such as Cloudsmith, Artifactory and nuget.org reject one that is
+// missing the OPC parts below:
+//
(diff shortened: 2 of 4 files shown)


YuKitsune:
### Code review

Found 2 issues:

1. `buildOpcParts` unconditionally adds the nuspec to OPC relationships and content types, but `shouldGenerateNuSpec` can return `false` (when no `--author`, `--title`, `--description`, etc. flags are set). In that case no `.nuspec` file is written to the archive, yet `_rels/.rels` still points at `/<id>.nuspec` and `[Content_Types].xml` still declares the `nuspec` extension. Strict OPC readers and feeds may reject the package.

https://github.com/OctopusDeploy/cli/blob/21f1011eca1ffcd3356fab3297de4e7ee2dfd48f/pkg/cmd/package/nuget/create/opc.go#L86-L112
https://github.com/OctopusDeploy/cli/blob/21f1011eca1ffcd3356fab3297de4e7ee2dfd48f/pkg/cmd/package/nuget/create/create.go#L114-L129

2. `defaultContentType` is set to `"application/octet"` but the correct MIME type is `"application/octet-stream"`. The comment says this is what NuGet itself writes, but NuGet and `dotnet pack` use `application/octet-stream`. `"application/octet"` is not a registered IANA MIME type.

https://github.com/OctopusDeploy/cli/blob/21f1011eca1ffcd3356fab3297de4e7ee2dfd48f/pkg/cmd/package/nuget/create/opc.go#L36-L38

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

YuKitsune:
I told Claude not to post the review as a comment, but it did anyway... I'll follow up with a manual review shortly.

YuKitsune:
Issue 2 that Claude identified is bogus. `application/octet` is correct for nuget packages, not `application/octet-stream`.

YuKitsune:
Issue 1 that Claude identified is pre-existing.
TL;DR: the `.nuspec` file isn't added unless Description, Title, Release Notes, or Author is provided. If you don't provide them, it'll still let you create the package, but it won't contain a nuspec file. `_rels/.rels` will be written with a reference to the nuspec file, even if it isn't created.

To fix it, you'd need to always create the nuspec file with a package ID and version.

Since it's pre-existing, I'll leave it up to you if you want to address that here or separately.

YuKitsune (approved):

PR #633, opened 2026-08-04
fix: stop task wait hanging on stdin and clearing the terminal line

### LINKED ISSUE #290: Moving an octocli command to the background clears the active line
## The bug
When I run octopus task wait it consistently prints out and clears the line. This is not what I would expect with --no-prompt. It should run, output new information as it gets it, and return 0 when completed. 

## Command to reproduce

```bash
octopus task wait $taskId --no-prompt -f basic -s $OCTOPUS_SPACE &
# outputs, then repeatedly clears the active line
```

## Outcome
The terminal becomes unusable with this task backgrounded due to the manual clearing. It should not ever clear if --no-prompt is selected as I'm explicitly not seeking an interactive experience, I just want the output and the command to end when there is a result. 

## Versions

**cli**: latest

**Octopus Server**: all

### LINKED ISSUE #526: octopus task wait cli command does not return if shelled out to from ruby
## The bug

Shelling out to `octopus task wait` in ruby with `Open3.popen3` never returns. Even when specifying `--no-prompt` and/or `--output-format json`. Even when the task has completed on the server. The calling ruby code will just flat-out never continue past the point where the call to `octopus task wait` was made.

## Command to reproduce

```ruby
require 'open3'
result = {messages: [], errors: [] }
command_name = 'task wait'

command_line = '/usr/local/bin/octopus task wait "ServerTasks-358388" --space "Spaces-1" --timeout 60 --no-prompt --output-format json'

Open3.popen3(command_line) do |stdin, stdout, stderr, wait_thr|
stdout.each do |line|
  line.chomp!
  if line.length.positive?
    trimmed_line = line.strip.squeeze("\t", ' ')
    result[:messages] << trimmed_line
  end
end

stderr.each do |line|
  line.chomp!
  if line.length.positive?
    trimmed_line = line.strip.squeeze("\t", ' ')
    result[:errors] << trimmed_line
    logger.error "#{trimmed_line}"
  end
end

exit_status = wait_thr.value
exit_message = ''
if exit_status.success?
  exit_message = "#{command_name} result: succeeded"
  result[:messages] << exit_message
  result[:overall_result] = :succeeded
  logger.info exit_message
else
  exit_message = "#{command_name} returned a failure code: #{exit_status.exitstatus}."
  result[:messages] << exit_message
  result[:overall_result] = :failed_unexpected_exit_code
  logger.warn exit_message
end


```

## Outcome

```
### crickets...

```

## Versions

**cli**: 1.8.0

**Octopus Server**: 2025.1.10012

**cli running on**: macOS Sequoia 15.5 (24F74) on m4 max macbook pro

NickJosevski:
Fixes #526
Fixes #290

Two faults in `task wait`, both triggered when not in interactive shell. 

## #526 — hangs forever when shelled out to

`task wait` read stdin unconditionally to pick up piped task IDs:

```go
taskIDs = append(taskIDs, util.ReadValuesFromPipe()...)
```

`IsCalledFromPipe()` is true whenever stdin is not a character device, and `scanner.Scan()` then blocks til writer closes pipe. 

### Testing
Piping still works: `echo ServerTasks-1 | octopus task wait` behaves as before.

## #290 — erases the terminal line when backgrounded

spinner runs  **before cobra parses any flags**. `--no-prompt` does not disable interactive mode until `PersistentPreRunE`  then decision is already made  and `--no-prompt` never suppressed the spinner. Backgrounded, it redraws every poll by erasing the current terminal line = unusable.

**Fix:** the round-tripper decides per request rather than at construction, once the flag has been applied.

### Testing

live instance, with a FIFO held open by a background writer to imitate `popen3`, timing the CLI process


YuKitsune (approved):
Tested against a local instance, seems to work!

PR #634, opened 2026-08-04
fix: make delete commands usable with --no-prompt

### LINKED ISSUE #490: `octopus deployment-target delete --no-prompt` doesn't work
## The bug

I'm trying to use the CLI to bulk-delete 400 deployment targets; this is exceedingly tedious to do by hand in the web portal

This powershell should get rid of every target in the system

```shell
octopus deployment-target list -f basic | %{ octopus deployment-target delete $_ --no-prompt }
```

However it just prints `prompt disabled` 400 times. 

To delete an individual target this should work, but also fails

## Command to reproduce

```
octopus deployment-target delete TARGETNAME --no-prompt 
```

## Outcome

`prompt-disabled` is printed. The deployment target is not deleted and there is no error mressage or indication of anything else that might have gone wrong.

Note: If I don't specify `--no-prompt` then the CLI interactively prompts me to confirm the deletion. This works and the target is deleted, but it is of no use when scripting to delete 400 things.

## Versions

**cli**: `2.1.5`

**Octopus Server**: `2025.2.933` (cloud preprod instance)

### LINKED ISSUE #530: package delete doesn't work with --no-prompt flag
## The bug

When using the octopus package delete command with the --no-prompt flag, i get the same error message when the -p flag is not used, even when it's specified. This also makes usage in GitHub Actions impossible. I suspect that the flag is allways active when using Actions.

## Command to reproduce

```
octopus package delete -p "Package" -v "Version" -y --no-prompt
```

## Outcome

No access right now same output as:

```
octopus package delete -v "Version" -y --no-prompt
```

## Versions

**cli**: 2.8.0

**Octopus Server**: Not sure

NickJosevski:
Fixes #490
Fixes #530

Two reports about `--no-prompt` on delete commands. Same symptom, different causes.

## #490 — `prompt disabled`, nothing deleted, no explanation

Deleting asks the user to retype the item's name, which cannot happen once prompting is off. `deployment-target delete <name> --no-prompt` therefore printed "prompt disabled" didn't work no hints

**Fix:** report which flag skips the confirmation.

```
before  prompt disabled
after   cannot confirm deletion of the deployment target while prompting is disabled;
        pass --confirm to delete it without confirmation
```

Handled in `DeleteWithConfirmation`, all delete commands covered.

*not* making `--no-prompt` imply confirmation of delete, don't want to silently destroy things. `-y --no-prompt` is the scripting path, and it works:

## #530 — `--package-id` and `--version` ignored under `--no-prompt`

Different mechanism. Those flags were consumed only inside `PromptMissing`, which `--no-prompt` skips entirely, so `deleteRun` fell through to demanding the positional ID:

```
$ octopus package delete -p "Package" -v "1.0.0" -y --no-prompt
package identifier is required but was not provided     # identical to omitting -p
```

**Fix:** resolve them regardless of prompting — nothing prompts while both are present. if value missing, say which flag is needed

## Testing

 
| What | Outcome |
|---|---|
| `target delete <name> --no-prompt` | actionable message, nothing deleted |
| `target delete <name> -y --no-prompt` | deleted |
| `package delete -p X -v Y -y --no-prompt` | deleted |
| `package delete --no-prompt` | names both flags |
| interactive `target delete <name>` | still prompts to retype the name |

### DIFF
--- a/pkg/cmd/package/delete/delete.go
+++ b/pkg/cmd/package/delete/delete.go
@@ -82,10 +82,13 @@ func NewCmdDelete(f factory.Factory) *cobra.Command {
 }
 
 func deleteRun(opts *DeleteOptions) error {
-	if !opts.NoPrompt {
-		if err := PromptMissing(opts); err != nil {
-			return err
-		}
+	// Resolve regardless of whether prompting is enabled. --package-id and
+	// --version identify the package just as well as the positional ID does,
+	// and running this only in interactive mode left them ignored under
+	// --no-prompt, which then rejected the command for want of an identifier.
+	// Nothing here prompts while both flags are supplied.
+	if err := PromptMissing(opts); err != nil {
+		return err
 	}
 
 	if opts.ID == "" {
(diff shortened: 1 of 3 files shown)


pkg/cmd/package/delete/delete.go
```diff
@@ -82,14 +82,22 @@ func NewCmdDelete(f factory.Factory) *cobra.Command {
 }
 
 func deleteRun(opts *DeleteOptions) error {
+	// --package-id and --version identify the package just as well as the
+	// positional ID does, so they are resolved whether or not prompting is
+	// enabled. Doing this inside PromptMissing left them ignored under
+	// --no-prompt, which then rejected the command for want of an identifier.
```
YuKitsune: Claude likes to leave comments like this, but they're often confusing and irrelevant outside the context of the PR.

```suggestion
	// enabled.
```

YuKitsune (approved):

PR #653, opened 2026-08-05
test: read stdout only while the asker goroutine is parked

NickJosevski:
## Problem

`TestDeployCreate_AskQuestions` runs `AskQuestions` in a goroutine and asserts on the buffer it writes to. Four subtests read that buffer straight after answering a question. Answering lets the goroutine run on and append the advanced-options summary, so the read races it.

Both outcomes turn up in CI: the buffer holds more than expected, or `stdout.Reset()` has already run and it holds nothing.

One of the four also had its arguments reversed, checking that a two-line heredoc contained the whole of stdout. That fails as soon as anything else is written.

## Change

Read the buffer only while the goroutine is parked on a question. Three assertions move below the next `ExpectQuestion`. The fourth compares exact contents, so it moves above the `AnswerWith` instead.

Corrected the reversed arguments. Dropped an `assert.Regexp(t, "", ...)`, which matched anything.

Forcing the race with a sleep in each window reproduces the CI failures on the old code, and none on the new.

### DIFF
--- a/pkg/cmd/release/deploy/deploy_test.go
+++ b/pkg/cmd/release/deploy/deploy_test.go
@@ -47,10 +47,6 @@ var now = func() time.Time {
 }
 var ctxWithFakeNow = context.WithValue(context.TODO(), constants.ContextKeyTimeNow, now)
 
-// AskQuestions runs in a goroutine here, so the stdout buffer is only stable
-// while that goroutine is parked on a question. Read it directly after an
-// ExpectQuestion returns; answering releases the goroutine to write more into
-// the same buffer, and a read taken after that races it.
 func TestDeployCreate_AskQuestions(t *testing.T) {
 	const spaceID = "Spaces-1"
 	const fireProjectID = "Projects-22"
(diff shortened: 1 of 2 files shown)


pkg/cmd/release/deploy/deploy_test.go
```diff
@@ -47,6 +47,10 @@ var now = func() time.Time {
 }
 var ctxWithFakeNow = context.WithValue(context.TODO(), constants.ContextKeyTimeNow, now)
 
+// AskQuestions runs in a goroutine here, so the stdout buffer is only stable
+// while that goroutine is parked on a question. Read it directly after an
+// ExpectQuestion returns; answering releases the goroutine to write more into
+// the same buffer, and a read taken after that races it.
```
YuKitsune: This comment seems like it would be better placed on `ExpectQuestion` itself. Might also help Claude do the assertions in the correct order in the future.

YuKitsune (approved):
It'd be nice if we could design away the race condition so we don't fall into this trap again, though I imagine that'll be a much bigger PR. Happy for this to merge as-is.

PR #622, opened 2026-08-04
fix: support colour output in CI via FORCE_COLOR/CLICOLOR_FORCE

NickJosevski:
Fixes #506

## Problem

Colour was gated solely on `term.IsTerminal(os.Stdout.Fd())`, which never resolves to true inside a CI job. CI systems that *do* render ANSI codes — GitHub Actions, GitLab CI — could therefore never show coloured output.

## Change

`pkg/output/color.go` now follows the conventions from [no-color.org](https://no-color.org) and [bixense.com/clicolors](https://bixense.com/clicolors):

| Condition | Result |
|---|---|
| `NO_COLOR` set to anything non-empty | colour off (highest precedence) |
| `CLICOLOR_FORCE` or `FORCE_COLOR` set to anything but `0` | colour on, even without a TTY |
| otherwise | existing terminal detection |

This is **opt-in**, so CI output stays plain unless asked for. Auto-enabling on a `CI` env var would inject escape codes into the logs of systems that don't render them, and would break anyone currently parsing CLI output in a pipeline.

## Verification

Unit tests cover the precedence matrix. Manually against a build:

```
$ octopus --help | cat -v | head -3
Work seamlessly with Octopus Deploy from the command line.
                ####

$ FORCE_COLOR=1 octopus --help | cat -v | head -3
Work seamlessly with Octopus Deploy from the command line.
^[[0;36m                ####
```

README documents the variables under Getting Started.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #659, opened 2026-08-05
chore: remove the colour demo workflow

NickJosevski:
Cleanup for #622. **Base is `nj/fix-506`, not `main`** — merge this into #622's branch once #622 has been reviewed, then #622 is ready for `main`.

## What this removes

`.github/workflows/colour-demo.yml`, the demo added in e06d7fe. It builds the CLI at three points in history — `main`, #622 as first reviewed, and #622 with the follow-up fixes — and prints `octopus --help` under a matrix of colour environment variables, so reviewers can see the before and after in a real CI log:

- [Run showing the original bug and the fix](https://github.com/OctopusDeploy/cli/actions/runs/30981377071)

It served two review conversations: the `FORCE_COLOR=0` behaviour YuKitsune raised, and the discovery that GitHub Actions resets colour at every line break, which meant the logo reached CI coloured but rendered white until the per-line fix in 8c95e27.

Once #622 merges, the three-way history comparison stops being meaningful and the workflow is dead weight.

## Why nothing is lost

Everything it demonstrated is asserted by unit tests in `pkg/output/color_test.go`:

| Behaviour | Test |
|---|---|
| Full precedence matrix, with **and** without a TTY | `TestIsColorEnabled` |
| Every exported helper honours `IsColorEnabled` | `TestColorHelpersHonourIsColorEnabled` |
| Multi-line text is coloured one line at a time | `TestMultiLineTextIsColouredPerLine` |
| Blank lines collect no stray escapes | `TestBlankLinesAreNotColoured` |

The workflow only ever ran on `workflow_dispatch` and on pushes to `nj/fix-506`, so it has never touched normal CI and removing it changes no other build.

## If you would rather keep it

Say so on #622 and close this instead. It would need the three-way build trimmed to just the current binary and the branch filter dropped, since neither survives the merge.


NickJosevski:
Superseded. The removal is now the last commit on #622 itself (ca8bccf), so it is a single self-contained PR rather than a stack.

The demo run is still viewable if anyone wants the before and after: https://github.com/OctopusDeploy/cli/actions/runs/30981377071

PR #681, opened 2026-08-12
fix: resolve the project in release progression allow and prevent

NickJosevski:
`octopus release progression allow --project X` panics.

Both commands only assign `selectedProject` when `--project` is absent, then use it regardless:

```go
if opts.Project.Value == "" {
    selectedProject, err = selectors.Project(...)   // only assigned when empty
}
opts.Project.Value = selectedProject.GetName()      // nil when --project was given
```

`PromptMissing` runs whenever prompting is enabled, so supplying `--project` interactively dereferences nil. The same shape makes `--version` panic on `selectedRelease.Version`.

`prevent` keeps its assignments inside the `if`, so it doesn't panic — but it hands the nil project to `SelectRelease`, which fails with `invalid parameter project` instead of listing releases.

## Fix

Look the project up when it was supplied, so the release prompt has a real project, and only assign the version on the path that prompts for it.

## Tests

First tests for either command. Both directions covered: `--project` supplied with the release prompted, and `--version` supplied with the project prompted.

Verified they fail without the fix — `allow` panics on `(*Project).GetName`, `prevent` hangs waiting for a project lookup that never happens.

Found while auditing `selectors.Project` call sites for #667, which deliberately left these two alone because adopting the shared helper there would change behaviour rather than just deduplicate. Independent of #609/#667 and mergeable on its own.

### DIFF
--- a/pkg/cmd/release/progression/prevent/prevent_test.go
+++ b/pkg/cmd/release/progression/prevent/prevent_test.go
@@ -1,0 +1,118 @@
+package prevent_test
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/AlecAivazis/survey/v2"
+	cmdRoot "github.com/OctopusDeploy/cli/pkg/cmd/root"
+	"github.com/OctopusDeploy/cli/pkg/question"
+	"github.com/OctopusDeploy/cli/test/fixtures"
+	"github.com/OctopusDeploy/cli/test/testutil"
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/defects"
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/projects"
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/releases"
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/resources"
+	"github.com/spf13/cobra"
+	"github.com/stretchr/testify/assert"
+)
+
+var rootResource = testutil.NewRootResource()
+
+func TestReleaseProgressionPrevent(t *testing.T) {
+	const spaceID = "Spaces-1"
+	const projectID = "Projects-22"
+	const releaseID = "Releases-9"
+
+	space1 := fixtures.NewSpace(spaceID, "Default Space")
+	fireProject := fixtures.NewProject(spaceID, projectID, "Fire Project", "Lifecycles-1", "ProjectGroups-1", "")
+
+	release := releases.NewRelease("Channels-1", projectID, "1.0")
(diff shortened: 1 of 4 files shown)


pkg/cmd/release/progression/prevent/prevent_test.go
```diff
@@ -0,0 +1,118 @@
...
+			api.ExpectRequest(t, "POST", "/api/Spaces-1/releases/Releases-9/defects").
+				RespondWith(&defects.Defect{Status: defects.DefectStatusUnresolved})
+
+			_, err := testutil.ReceivePair(cmdReceiver)
+			assert.Nil(t, err)
+			assert.Contains(t, stdOut.String(), "Successfully prevented progression for release 1.0")
+			assert.Equal(t, "", stdErr.String())
+		}},
+
+		{"prevent prompts for the project when the version is supplied", func(t *testing.T, api *testutil.MockHttpServer, qa *testutil.AskMocker, rootCmd *cobra.Command, stdOut *bytes.Buffer, stdErr *bytes.Buffer) {
```
YuKitsune:
```suggestion
		{"prevent prompts for the release when the version is supplied", func(t *testing.T, api *testutil.MockHttpServer, qa *testutil.AskMocker, rootCmd *cobra.Command, stdOut *bytes.Buffer, stdErr *bytes.Buffer) {
```

pkg/cmd/release/progression/prevent/prevent_test.go
```diff
@@ -0,0 +1,118 @@
...
+	release.ID = releaseID
+	release.SpaceID = spaceID
+
+	tests := []struct {
+		name string
+		run  func(t *testing.T, api *testutil.MockHttpServer, qa *testutil.AskMocker, rootCmd *cobra.Command, stdOut *bytes.Buffer, stdErr *bytes.Buffer)
+	}{
+		// Regression: --project supplied interactively used to leave selectedProject nil, which
+		// was then handed to SelectRelease and failed with "invalid parameter project".
+		{"prevent prompts for the release when the project is supplied", func(t *testing.T, api *testutil.MockHttpServer, qa *testutil.AskMocker, rootCmd *cobra.Command, stdOut *bytes.Buffer, stdErr *bytes.Buffer) {
```
YuKitsune:
```suggestion
		{"prevent prompts for the project when the project is supplied", func(t *testing.T, api *testutil.MockHttpServer, qa *testutil.AskMocker, rootCmd *cobra.Command, stdOut *bytes.Buffer, stdErr *bytes.Buffer) {
```

YuKitsune (approved):
Two issues with test case names, otherwise no red flags.

PR #682, opened 2026-08-12
refactor: address channel command review feedback

NickJosevski:
Actions the review feedback on #609. Stacked on that branch.

- `delete`: validate flags up front in automation mode, as `release delete` does
- `delete`: use `*cmd.Dependencies` instead of hand-rolled `Client`/`Ask`/`Out`/`NoPrompt` fields
- `delete`: drop the Config-as-Code warning — no other version-controlled command has one
- `shared.ResolveChannel`: prompt when no channel is given, fail in automation mode, mirroring `selectors.ResolveProject`. Removes the hand-rolled prompt in `delete`
- `list`: show `Inherited from project` when a channel has no lifecycle of its own
- `list`: capitalise the `--filter` description

### DIFF
--- a/pkg/cmd/channel/delete/delete.go
+++ b/pkg/cmd/channel/delete/delete.go
@@ -2,21 +2,15 @@ package delete
 
 import (
 	"errors"
-	"fmt"
 
-	"github.com/AlecAivazis/survey/v2"
 	"github.com/MakeNowJust/heredoc/v2"
-	"github.com/OctopusDeploy/cli/pkg/apiclient"
+	"github.com/OctopusDeploy/cli/pkg/cmd"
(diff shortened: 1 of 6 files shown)


NickJosevski:
Closing as absorbed. `3c9a310` on the base branch (#609) is a refined redo of the same change — `git range-diff` matches it as the same patch, with better naming and comments on top. Rebasing this branch onto the updated base resolves every conflict in favour of the base and leaves an empty commit, so there is nothing here to land.

The one thing this branch had that `3c9a310` didn't was reusing `selectors.Channel` rather than hand-rolling `promptForChannel`. That's now applied directly to #609 in `e3103f7`, which also picks up the selector's behaviour of taking the only channel when a project has exactly one.

PR #689, opened 2026-08-18
feat: add a group filter to project list

NickJosevski:
Adds an optional group filter to `octopus project list`:

```
octopus project list --group 'Default Project Group'
octopus project ls -g ProjectGroups-1
```

### Why

There was no way to narrow `project list` to a single project group. The nearest thing was `project-group view`, which shows group metadata with the projects nested inside it — fine for a look, awkward when the projects are what you actually want.

This is the alternative shape proposed on #388, which added a `project-group projects list` command instead. Putting the filter on `project list` is more discoverable (people looking for projects reach for `project list`), matches the `-g/--group` flag `project create` and `project clone` already use, and avoids a three-level `project-group projects list` whose middle noun would exist to hold a single verb. Credit for spotting the gap goes to @nate-christensen.

### Behaviour

- No flag: unchanged — `Projects.GetAll()`, same output and columns as before.
- With a group: resolved by name **or** ID, then listed from that group's own `Projects` link, which the server pages for us rather than us fetching every project and discarding most of them. (The projects collection has no `projectGroupId` filter, so this is the available route either way.)
- Unknown group: `cannot find a project group with name or ID of 'X'`, matching the wording in the `delete` commands, rather than the SDK's bare `cannot find the item`. Other lookup errors pass through untouched.

The selection logic sits in a small `getProjects` function taking callbacks, so it is unit-testable without a factory mock — five tests cover no-filter, filter, unknown group, nil group and unrelated-error paths.

### Verification

Build, vet, gofmt and tests clean. Against a local instance: created a throwaway group and project, confirmed the filter discriminates by both name and ID (16 projects total, 1 in the temp group, 15 in the default), then removed them. Also diffed the raw field sets returned by `/projects` and `/projectgroups/X/projects` — identical, `ProjectTags` included, so the `TAGS` column behaves the same on either path.

### Decision

Resolved: we're going with this shape. #388 proposed the same capability as a
`project-group projects list` command; it has been closed in favour of this PR, with
thanks to @nate-christensen for spotting the gap. The reasoning, plus two flag
registration bugs from that branch worth knowing about regardless, is written up in
https://github.com/OctopusDeploy/cli/pull/388#issuecomment-5265730529.

Out of draft and ready for review.


pkg/cmd/project/list/list.go
```diff
@@ -66,3 +100,31 @@ func listRun(cmd *cobra.Command, f factory.Factory) error {
...
+	if err != nil {
+		if errors.Is(err, services.ErrItemNotFound) {
+			return nil, fmt.Errorf("cannot find a project group with name or ID of '%s'", group)
+		}
+		return nil, err
+	}
+
+	if projectGroup == nil {
+		return nil, fmt.Errorf("cannot find a project group with name or ID of '%s'", group)
+	}
```
NickJosevski, line 124: The "cannot find a project group" message is built in two places, and the `projectGroup == nil` branch is unreachable with the real callback (`ProjectGroups.GetByIDOrName` never returns `nil, nil` — it either returns a group or `services.ErrItemNotFound` from `GetByName`). Collapsing them keeps one copy of the message while staying defensive:

```suggestion
	projectGroup, err := getProjectGroup(group)
	if err != nil && !errors.Is(err, services.ErrItemNotFound) {
		return nil, err
	}
	if err != nil || projectGroup == nil {
		return nil, fmt.Errorf("cannot find a project group with name or ID of '%s'", group)
	}
```

pkg/cmd/project/list/list_test.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/projects"
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/services"
+	"github.com/stretchr/testify/assert"
+)
+
+func newProject(name string) *projects.Project {
+	return projects.NewProject(name, "Lifecycles-1", "ProjectGroups-1")
+}
+
+func TestGetProjects_WithoutGroupListsEverything(t *testing.T) {
```
NickJosevski, line 17: These tests cover the extracted `getProjects` helper nicely, but nothing exercises the cobra wiring: that `--group`/`-g` is registered, that `listRun` passes the flag value through, or the output mapping. Every sibling list-command test (`pkg/cmd/channel/list/list_test.go`, `release/list`, `worker/list`, `target/list`, `package/list`) drives the command end-to-end through `testutil.MockHttpServer` and `rootCmd.ExecuteC()`, which would catch a regression in the flag registration or run path that these unit tests can't. Worth adding one end-to-end case in that established harness (e.g. `project list -g ProjectGroups-1 -f json` expecting `GET /api/Spaces-1/projectgroups/ProjectGroups-1` then the group's projects link).

pkg/cmd/project/list/list.go
NickJosevski: For the record, since this thread was auto-resolved by the *Apply suggestion* button rather than by the change itself: the collapse actually landed by hand in 74ad1b4. The apply commit (35a5361) re-ran this same suggestion against code that 74ad1b4 had already collapsed, matched the wrong line range, and deleted the adjacent `if group == "" { return getAllProjects() }` guard along with the function's closing brace — which broke the build. Reverted in 42591e5.

The substance of this comment is in place, so leaving the thread resolved.

pkg/cmd/project/list/list_test.go
NickJosevski: Done in e25bfdf: a `MockHttpServer` / `rootCmd.ExecuteC()` suite in `list_e2e_test.go`, built in the same shape as `pkg/cmd/package/list`. It covers the unfiltered path, `--group` by ID, the `-g` shorthand with a group name in json (ID lookup misses, partial name lookup resolves), and the unknown-group error.

Worth noting these earned their keep immediately — they were the tests taken down by the broken apply commit, and they pass again as of 42591e5.

PR #694, opened 2026-08-19
feat: toggle the enabled state of deployment targets

### LINKED ISSUE #311: Enhancement: Add support for toggling enabled state of deployment targets
## The bug
It appears there is no support for toggling the enabled state of deployment targets.
The deployment target enable state could be set both upon deployment target creation and through a new deployment target update command.

## Command to reproduce
octopus deployment-target azure-web-app create [flags]

## Outcome
Deployment target enable with no facility to change the enabled state through the CLI.

## Versions
2.1.0

NickJosevski:
Fixes #311

## What changed

**New commands** — `octopus deployment-target enable [<name> | <id>]` and `octopus deployment-target disable [<name> | <id>]`.

- Resolves the target with `Machines.GetByIdentifier`, flips `IsDisabled`, and saves via `machines.Update`.
- If the target is already in the requested state it prints a message and makes no update call.
- With no positional argument (and prompting enabled) it prompts for the target, offering only targets not already in the requested state. Under `--no-prompt` a missing identifier is an error.
- Shared logic lives in `pkg/cmd/target/shared/disabledstate.go`; `enable`/`disable` are thin cobra wrappers over it.

**New create flag** — `--disabled` on every deployment-target `create` command (`azure-web-app`, `cloud-region`, `kubernetes`, `listening-tentacle`, `ssh`), registered from a shared `machinescommon.CreateTargetDisabledFlags` and included in the generated automation command. It is flag-only: it deliberately does not add a prompt to the interactive create flow.

**Test support** — `testutil.NewRootResource()` now publishes the `Machines` link so command tests can exercise the machines endpoints.

## Decisions

These were open questions in an earlier revision. They are settled, each on an existing CLI convention.

- **Dedicated `enable`/`disable` subcommands, not the `update` command the issue suggests.** `project` and `tenant` already ship `enable|disable`, and the CLI has no resource-level `update` command anywhere — the only two are `project variables update` and `tenant variables update`. A future `deployment-target update` can still cover the remaining editable fields; enable/disable stays the ergonomic shortcut.
- **`--disabled`, not `--enabled=false` or `--disable`.** No bool flag in the CLI defaults to `true`, so `--enabled=false` would be the first; and flag names are states (`--prompted`, `--unscoped`, `--default`), not verbs. `--disabled` also matches the `IsDisabled` API field and the `IS DISABLED` column `tenant list` already prints.
- **Workers get the same treatment** — `worker enable|disable` plus `--disabled` on the worker create commands, with the shared helper generalised into `pkg/machinescommon`: #731, stacked on this branch.
- **`list`/`view` surface the disabled state** — following `tenant list`/`tenant view`, for both targets and workers: #732, stacked on this branch.

#731 and #732 are independent of each other and touch no files in common.

## Test evidence

New tests:

- `pkg/cmd/target/enable/enable_test.go`, `pkg/cmd/target/disable/disable_test.go` — end-to-end through the root command against the mock HTTP server, in the style of `release/progression/prevent`: named target, already-in-state short circuit, and the interactive prompt path. The PUT body is asserted to carry the flipped `IsDisabled`.
- `pkg/cmd/target/shared/disabledstate_test.go` — prompting behaviour (identifier supplied vs not, and the enable/disable wording).
- `pkg/cmd/target/target_test.go` — the enable/disable subcommands are registered and every target `create` command exposes `--disabled`.
- `test/integration/target_test.go` — against a real server: `--disabled` on two endpoint types, and that a toggle leaves the target's other settings untouched.

Results from the worktree:

- `go build ./...` — clean.
- `go test ./pkg/...` — all packages pass, no failures. (`gofmt`/`go vet` report only pre-existing issues in files this PR does not touch.)

No live server is required by the unit tests.


pkg/cmd/target/shared/disabledstate.go
```diff
@@ -0,0 +1,95 @@
...
+	_, _ = fmt.Fprintf(opts.Out, "Successfully %s deployment target '%s' %s.\n", state, target.Name, output.Dimf("(%s)", target.GetID()))
+	return nil
+}
+
+func PromptMissingTarget(opts *SetDisabledStateOptions, isDisabled bool) error {
+	if opts.IdOrName != "" {
+		return nil
+	}
+
+	selectedTarget, err := selectors.Select(
```
NickJosevski: `selectors.Select` auto-returns the item when the list has exactly one entry (pkg/question/selectors/selectors.go:48), so in a space with a single deployment target, a bare interactive `octopus deployment-target disable` disables that target immediately with no prompt or confirmation at all. Note `tenant disable` goes through `selectors.ByName` → `question.SelectMap`, which always asks even for one item — so this command is both silently mutating and inconsistent with the existing enable/disable pattern.

pkg/cmd/target/shared/disabledstate.go
```diff
@@ -0,0 +1,95 @@
...
+		opts.Ask,
+		fmt.Sprintf("Select the deployment target you wish to %s:", actionDescription(isDisabled)),
+		opts.GetTargetsCallback,
+		func(target *machines.DeploymentTarget) string { return target.Name })
+	if err != nil {
+		return err
+	}
+
+	opts.IdOrName = selectedTarget.GetID()
```
NickJosevski, line 114: The prompt path fetches every machine in the space (`take=2147483647`) and the user picks a full `*machines.DeploymentTarget`, but only its ID is kept — `SetDisabledState` then re-fetches the same object via `GetByIdentifier` (an extra GET, or two when the identifier falls through to the name lookup). Carrying the selected target (e.g. a `target *machines.DeploymentTarget` field on the options, with `GetByIdentifier` only for the CLI-supplied identifier) drops a redundant API round-trip.

pkg/cmd/target/shared/disabledstate.go
```diff
@@ -0,0 +1,95 @@
...
+
+func NewSetDisabledStateOptions(args []string, dependencies *cmd.Dependencies) *SetDisabledStateOptions {
+	idOrName := ""
+	if len(args) > 0 {
+		idOrName = args[0]
+	}
+
+	return &SetDisabledStateOptions{
+		Dependencies:      dependencies,
+		GetTargetsOptions: NewGetTargetsOptionsForAllTargets(dependencies),
```
NickJosevski: The selection list is unfiltered, so `disable` offers already-disabled targets and `enable` offers already-enabled ones; picking one ends in the "is already disabled/enabled" no-op after the whole interactive flow. For `enable`, the machines endpoint supports server-side filtering (`MachinesQuery{IsDisabled: true}` — the query field is `omitempty` so only the true case can be expressed); for `disable` the list could be filtered client-side on `!target.IsDisabled`. Filtering would also make the single-item auto-select less surprising.

pkg/cmd/target/listening-tentacle/create/create.go
```diff
@@ -147,14 +150,16 @@ func createRun(opts *CreateOptions) error {
...
 	}
 
+	deploymentTarget.IsDisabled = opts.Disabled.Value
+
 	createdTarget, err := opts.Client.Machines.Add(deploymentTarget)
 	if err != nil {
 		return err
 	}
 
 	fmt.Fprintf(opts.Out, "Successfully created listening tenatcle '%s'.\n", deploymentTarget.Name)
```
NickJosevski: Pre-existing, but since this function is touched by the PR: user-visible typo "tenatcle" in the success message.

```suggestion
	fmt.Fprintf(opts.Out, "Successfully created listening tentacle '%s'.\n", deploymentTarget.Name)
```

pkg/cmd/target/disable/disable_test.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/OctopusDeploy/go-octopusdeploy/v2/pkg/resources"
+	"github.com/spf13/cobra"
+	"github.com/stretchr/testify/assert"
+)
+
+var rootResource = testutil.NewRootResource()
+
+const spaceID = "Spaces-1"
+
+func newTarget(id string, name string, isDisabled bool) *machines.DeploymentTarget {
```
NickJosevski: `newTarget`, `rootResource`, `spaceID`, and the whole table-test scaffold here are byte-for-byte duplicates of pkg/cmd/target/enable/enable_test.go. A shared helper (e.g. a deployment-target builder in test/fixtures next to `fixtures.NewSpace`) would keep the two files from drifting apart.

pkg/cmd/target/shared/disabledstate.go
NickJosevski: Actioned in 02ae422. The finding holds — `selectors.Select` returns `items[0]` without asking when `len(items) == 1` (pkg/question/selectors/selectors.go:48-50), so a single-target space got no prompt.

`PromptMissingTarget` now calls `question.SelectMap` directly instead of going through `selectors.Select`, which lines it up with the `tenant enable`/`disable` behaviour: the list is always presented, one item or twenty. `SelectMap` is what `selectors.Select` falls through to anyway, so the only thing dropped is the auto-select shortcut. There's a comment at the call site saying why, so nobody reinstates `selectors.Select` for consistency later.

Covered by `TestPromptMissingTarget_AsksEvenWhenThereIsOnlyOneTarget` in pkg/cmd/target/shared/disabledstate_test.go, which feeds a one-target callback and asserts the select prompt is still asked (the mock asker's `checkRemainingPrompts` fails if it isn't).

pkg/cmd/target/shared/disabledstate.go
NickJosevski: Actioned in 7627672. `SetDisabledStateOptions` gained a `Target *machines.DeploymentTarget` field; `PromptMissingTarget` stores the object the user picked, and `SetDisabledState` only falls back to `GetByIdentifier` when `Target` is nil — i.e. when the identifier came from the command line. `IdOrName` is still set from the selection so the non-prompt path and any caller reading it keep working.

The prompt path is now list + PUT with nothing in between. That's asserted structurally rather than by a counter: in the `prompts for the target when none was supplied` cases in pkg/cmd/target/disable/disable_test.go and pkg/cmd/target/enable/enable_test.go, the mock server expects `GET /machines?take=...` then `PUT /machines/Machines-200` directly — `MockHttpServer` fails on an unexpected request, so a re-added `GET /machines/Machines-200` would break the test.

Residual: the list itself is still `take=2147483647` over every machine in the space. That's `GetTargetsOptions` as `target list` and `target delete` already use it, so I left it alone; narrowing it would be a change to the shared helper rather than to this command.

pkg/cmd/target/shared/disabledstate.go
NickJosevski: Actioned in 62be612, using both halves of the suggestion.

`NewSetDisabledStateOptions` builds `machines.MachinesQuery{IsDisabled: !disabled}`, so `enable` sends `isDisabled=true` and `disable` sends nothing (the `omitempty` you noted). `PromptMissingTarget` then filters the returned list on `target.IsDisabled != opts.Disabled` regardless — the server-side filter only covers the enable case, and I didn't want the disable case relying on a filter the query type can't express. When nothing is left after filtering it errors with `no deployment targets to <enable|disable> were found` rather than dropping the user into an empty select.

Tests: `does not offer targets which are already disabled` in pkg/cmd/target/disable/disable_test.go returns one enabled and one disabled target and asserts the prompt options are `["web-server"]` only; the enable equivalent asserts the request URL is `/api/Spaces-1/machines?isDisabled=true&take=2147483647`, which pins the server-side filter. `TestPromptMissingTarget_ErrorsWhenNoTargetIsInTheOppositeState` covers the empty case.

The "is already disabled/enabled" no-op message is kept — it's still reachable via an explicit identifier with `--no-prompt`.

pkg/cmd/target/listening-tentacle/create/create.go
NickJosevski: Actioned in a3d40fb — took the suggestion as written. `grep -rn tenatcle` over the repo now returns nothing, so that was the only occurrence.

pkg/cmd/target/disable/disable_test.go
NickJosevski: Actioned in 314eafc. `newTarget` is gone from all three test files and replaced by `fixtures.NewDeploymentTarget(spaceID, targetID, name, isDisabled)` in test/fixtures/projects.go, next to `NewSpace`/`NewTenant` as suggested. There were actually three copies, not two — pkg/cmd/target/shared/disabledstate_test.go had a near-identical one as well; that's now on the shared fixture too.

Residual, deliberate: `var rootResource = testutil.NewRootResource()` and `const spaceID = "Spaces-1"` are still declared in each test package. They're package-level declarations rather than helpers, and the same pair appears in 29 test files across the repo, so hoisting them is a repo-wide change rather than something this PR should introduce a precedent for. The table-test scaffold itself (the `tests []struct{name, run}` slice and the `for _, test := range tests` runner) is likewise the standard shape used by every command test here — I kept it rather than abstracting it, since the two files' cases genuinely differ now that `disable` has the extra filtering case.

Issue #604, opened 2026-05-18
deployment target commands crash for AWS ECS targets


benPearce1:
## The bug

Running `deployment-target view` or `deployment-target list` where the target is AWS ECS will cause the CLI to crash

## Command to reproduce

```
octopus deployment-target view "aws ecs" -f basic -s default
```

## Outcome

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x18 pc=0x10341ee9c]

goroutine 1 [running]:
github.com/OctopusDeploy/cli/pkg/cmd/target/view.getDeploymentTargetAsBasic(0x1400018c8a0, 0x140002ea000)
        /Users/benpearce/git/cli/pkg/cmd/target/view/view.go:219 +0x38c
github.com/OctopusDeploy/cli/pkg/cmd/target/view.ViewRun.func3(0x16cf53107?)
        /Users/benpearce/git/cli/pkg/cmd/target/view/view.go:58 +0x28
github.com/OctopusDeploy/cli/pkg/output.PrintResource[...](0x140002ea000?, 0x140002f4008, {0x140001867c0, {{0x140002f65a0, 0x9, 0x9}, 0x140001867d0}, 0x140001867e0})
        /Users/benpearce/git/cli/pkg/output/print_resource.go:40 +0xb0
github.com/OctopusDeploy/cli/pkg/cmd/target/view.ViewRun(0x1400018c8a0)
        /Users/benpearce/git/cli/pkg/cmd/target/view/view.go:47 +0x1d4
github.com/OctopusDeploy/cli/pkg/cmd/target/view.NewCmdView.func1(0x140002f4008, {0x1400018ecd0, 0x1, 0x10348d82b?})
        /Users/benpearce/git/cli/pkg/cmd/target/view/view.go:32 +0xc0
github.com/spf13/cobra.(*Command).execute(0x140002f4008, {0x1400018ec80, 0x5, 0x5})
        /Users/benpearce/go/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:916 +0x5e0
github.com/spf13/cobra.(*Command).ExecuteC(0x1400029a008)
        /Users/benpearce/go/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:1044 +0x2ec
github.com/spf13/cobra.(*Command).Execute(0x103775d30?)
[...]

PR #701, opened 2026-08-19
fix: stop worker commands crashing on unreported endpoint details

NickJosevski:
Follow-up to #690, which called this out as out of scope. **Stacked on #690** — based on `nj/fix-604`, so review that one first; the base retargets to `main` once it merges.

## Problem

#690 fixed a class of crash for deployment targets: the CLI dereferenced endpoint fields the API is free to omit. The worker commands have the same unguarded reads. No issue has been filed for these — they were found while fixing #604.

Unlike the ECS case, a worker's communication style is always one the SDK models, so the nil-endpoint path is theoretical for them. These two are not:

- **`worker list` crashed for a Tentacle the server has never contacted.** `getVersion` dereferenced `TentacleVersionDetails`, which is `omitempty` and absent until the first health check. Like `deployment-target list`, it reads every worker, so one such worker took down the whole listing.
- **The type-specific views crashed when pointed at the wrong type.** `worker listening-tentacle view`, `worker polling-tentacle view` and `worker ssh view` each asserted the endpoint to their own type unchecked.

Also guarded, for the same reason as in #690: every read of the communication style in `worker list`, `worker view`, the three type-filtered lists, `worker-pool view` and `worker-pool static view`.

## Change

Workers need exactly the helpers #690 added for targets, so `GetCommunicationStyle`, `FormatUri`, `FormatTentacleVersion` and `EndpointAs` move from `pkg/cmd/target/shared` into `pkg/machinescommon` — the package the target, worker and worker pool commands already share. Nothing about their behaviour changes; the diff in `pkg/cmd/target/**` is the move and the call-site rename.

Two things collapse while passing through:

- Naming a communication style was about to become a third near-copy of the same map-lookup-with-fallback, so it is now one function, `DescribeCommunicationStyle`.
- `EndpointAs`'s mismatch message is now `this worker is not of type SSH` rather than `this deployment target is not a SSH deployment target` — one wording that works for both machine kinds and doesn't produce "a SSH".

`worker list`'s per-endpoint helpers now switch on the endpoint's Go type rather than its communication style string, which is what made the unchecked assertions possible in the first place.

## Behaviour changes worth a look

- A Tentacle with no version reports `Unknown` instead of crashing, matching `deployment-target view`.
- A `KubernetesTentacle` worker's JSON `Type` was `""` (a silent map miss) and is now `KubernetesTentacle`. Neither map in `machinescommon/comms.go` has an entry for that style — adding proper display names for it looks worthwhile but is a naming call I'd rather leave to someone who knows what the product calls it. The same gap shows in `deployment-target list` on a live server, where Kubernetes agents list as `KubernetesTentacle`.

## Verification

Both crashes reproduce on the pre-change code: copying `getVersion` verbatim into a test and calling it with a never-health-checked Listening Tentacle panics, as does calling it with no endpoint. Unit tests cover every function that reads the endpoint, including the nil, typed-nil and wrong-type cases. Full `go test ./pkg/...` passes.

## Why draft

Nothing here is user-reported, and it is stacked on an unmerged PR. Happy to split the `machinescommon` move out if reviewing the two together is awkward.


NickJosevski:
Superseded by #706, which is the same branch and commit based on #705 instead, so its diff is only the worker fixes. GitHub would not retarget this one once it had been pulled into a stack.

PR #705, opened 2026-08-19
refactor: move the endpoint read helpers into machinescommon

NickJosevski:
Split out of the worker fixes (#706) so the mechanical move is reviewable on its own. **Stacked on #690** — review that first; the base retargets to `main` as the stack merges. #706 sits on top of this and contains only the worker fixes.

## Why

#690 added four helpers to `pkg/cmd/target/shared` to guard reads of a deployment target's endpoint: `GetCommunicationStyle`, `FormatUri`, `FormatTentacleVersion` and `EndpointAs`. The worker and worker pool commands have the same unguarded reads (#706) and need the same helpers, so they move to `pkg/machinescommon` — the package the target, worker and worker pool commands already share.

## Two things collapse on the way through

- **Naming a communication style** was a map lookup with two fallbacks — friendly name, else the raw style, else `Unknown`. #690 wrote it once for `deployment-target list`, `deployment-target view` had its own variant, and #706 was about to add a third and fourth for workers. It is now one function, `DescribeCommunicationStyle(endpoint, names)`, taking whichever of the two maps in `comms.go` the caller wants.
- **`EndpointAs` takes the machine kind as a noun.** #690's message was `this deployment target is not a Listening Tentacle deployment target`, which repeats the noun and, for SSH, produces "a SSH". It now reads `this deployment target is not of type Listening Tentacle` / `this worker is not of type SSH` — one wording that works for both kinds and has no article to get wrong.

`GetCommunicationStyle` also changes shape: it took a `*machines.DeploymentTarget`, and now takes an `IEndpoint`, since a worker isn't a target. Callers pass `target.Endpoint` / `worker.Endpoint`, and the nil-target guard becomes a nil-endpoint guard — which is the case that actually matters, and which `machines.IsNil` handles for typed nils too.

## Verification

`go build ./...` and `go test ./pkg/...` pass on this commit alone, not just with #706 on top. The tests that moved with the helpers now live in `pkg/machinescommon/endpoint_test.go`, covering the nil, typed-nil, wrong-type and unnamed-style cases.

Checked command by command against a live instance, comparing this branch's binary with #690's: `deployment-target list` (table and json) and `deployment-target view` (basic and json, across ECS, Kubernetes, SSH and Tentacle targets) are identical, as are the type-specific views when pointed at a matching target.

**The one intended output change** is the mismatch message, exactly as described above:

```
#690:  this deployment target is not a Listening Tentacle deployment target
here:  this deployment target is not of type Listening Tentacle
```

If you would rather keep #690's wording, say so and I will drop that part — it is independent of the move.


pkg/machinescommon/endpoint.go
```diff
@@ -0,0 +1,69 @@
...
+		return ""
+	}
+
+	return endpoint.GetCommunicationStyle()
+}
+
+// Names a communication style for display: its friendly name where we have one,
+// the raw style where we do not, and "Unknown" when there is no endpoint to
+// read a style from at all.
+func DescribeCommunicationStyle(endpoint machines.IEndpoint, names map[string]string) string {
```
YuKitsune, line 31: Is `names` ever _not_ `machinescommon.CommunicationStyleToDescriptionMap`? Wondering if we can drop the parameter here.

YuKitsune (approved):
One possible bit to clean up, otherwise happy.

pkg/machinescommon/endpoint.go
NickJosevski: needed in #706, worker list calls it twice, with different maps:

list.go:77  Type: describeWorkerType(item.Endpoint)   → CommunicationStyleToDeploymentTargetTypeMap   (JSON)
list.go:92        describeWorkerStyle(item.Endpoint)  → CommunicationStyleToDescriptionMap            (table)

NickJosevski: this got messy with all the PR splits

PR #717, opened 2026-09-01
Report AWS ECS target detail now the SDK deserialises the endpoint

enf0rc3:
Unit tests have been failing on `main` since the go-octopusdeploy bump to v2.116.0 in #708. That release added `AwsEcsClusterEndpoint`, so AWS ECS deployment targets no longer arrive with a nil endpoint, and five tests asserting the old unknown-type fallbacks started failing.

The target commands now name the type as "AWS ECS Cluster", show the cluster and region as endpoint details, and resolve the default worker pool. Tests that covered the nil-endpoint path now use a target with no endpoint at all, so the #604 regression stays guarded.

Verified with `go test ./...` from `./pkg`: the three target packages that fail on `main` pass on this branch, and no other package changed state.

### DIFF
--- a/pkg/cmd/target/list/list_test.go
+++ b/pkg/cmd/target/list/list_test.go
@@ -16,12 +16,22 @@ func TestDescribeTargetType_KnownStyle(t *testing.T) {
 	assert.Equal(t, "Listening Tentacle", describeTargetType(target))
 }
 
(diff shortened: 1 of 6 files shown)


borland (approved):
LGTM

PR #726, opened 2026-09-08
fix: stop login --ignore-ssl-errors panicking on a wrapped transport

### LINKED ISSUE #725: octopus login --ignore-ssl-errors panics: interface conversion on SpinnerRoundTripper
### Version

2.22.0 and later (verified on 2.24.0 and 2.25.0; 2.21.4 is unaffected)

### What happened

`octopus login --ignore-ssl-errors` panics whenever the CLI has a server/API key available from its config or environment:

```
$ CI=true OCTOPUS_URL=http://localhost:8080 OCTOPUS_API_KEY=API-XXXX \
    octopus login --server http://localhost:8080 --api-key API-XXXX --ignore-ssl-errors --no-prompt --output-format json
panic: interface conversion: http.RoundTripper is *apiclient.SpinnerRoundTripper, not *http.Transport

goroutine 1 [running]:
github.com/OctopusDeploy/cli/pkg/cmd/login.loginRun(...)
	/home/runner/work/cli/cli/pkg/cmd/login/login.go:134 +0x33b
```

Dropping `--ignore-ssl-errors` from the same command works.

### Cause

`pkg/cmd/login/login.go:134` type-asserts the client's transport:

```go
httpClient.Transport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
```

Until 2.21.4 the spinner round tripper was only installed in interactive mode, so non-interactive runs got a `*http.Client` with a nil transport and the assertion held. `NewClientFactoryFromConfig` now installs it unconditionally (bafc85c5, released in 2.22.0), so the transport is a `*apiclient.SpinnerRoundTripper` and the assertion panics.

### Impact

This is not limited to people calling the CLI directly. The Octopus Jenkins plugin runs `octopus login --ignore-ssl-errors` whenever "Ignore SSL Errors" is configured on the server, so every build for those users crashes on CLI 2.22.0+. It is currently breaking that plugin's CI: https://github.com/OctopusDeploy/octopus-jenkins-plugin/actions/runs/34188365972/job/101941172315

### Steps to reproduce

1. `export OCTOPUS_URL=... OCTOPUS_API_KEY=...` (or log in once so a config file exists)
2. `octopus login --server <url> --api-key <key> --ignore-ssl-errors --no-prompt`

### Note

`NewClientFactoryFromConfig` also sets `http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: true}` unconditionally (`pkg/apiclient/client_factory.go:125`), so certificate verification is disabled for every config-backed command regardless of the flag. That looks unintended and is worth a separate look.

flin-8:
Fixes #725.
Fixes [SF-3525: Octopus CLI crash](https://linear.app/octopus/issue/SF-3525/octopus-cli-crash)

`login --ignore-ssl-errors` type-asserted the client's transport to `*http.Transport`. Since #633 the client factory always installs the spinner round tripper (it was interactive-only before, so non-interactive runs had a nil transport and the assertion held), so that assertion panics for anyone with a server and API key already in config or environment — including every Octopus Jenkins plugin build with "Ignore SSL Errors" set.

`skipTlsVerification` walks past the spinner and applies the setting to the transport underneath. A transport it does not recognise returns an error, rather than quietly leaving verification on after the caller asked for the opposite. Everything else, including the existing nil-transport handling, is unchanged.

The login test panics without the fix — the mock factory's transport is not an `*http.Transport` either, so it now reports the unsupported transport instead. Internal tests cover the spinner path and the error.

Thanks to #727, which fixed the same bug independently; the error on unrecognised transports is taken from it.


pkg/cmd/login/login_internal_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+	skipTlsVerification(transport)
+
+	assert.True(t, transport.TLSClientConfig.InsecureSkipVerify)
+}
+
+// The transport the client factory hands out is always spinner-wrapped, which is
+// what the old type assertion tripped over.
+func TestSkipTlsVerification_ThroughTheSpinner(t *testing.T) {
+	spinner := apiclient.NewSpinnerRoundTripper(nil)
```
NickJosevski, line 21: @flin-8 with the session I had going with Claude that spun up context from when we first started to fix the original driver for the changes in this space, on this it brings up:

`sets Next: http.DefaultTransport, the process-wide *http.Transport every client without its own uses.`

pkg/cmd/login/login.go
```diff
@@ -421,3 +421,14 @@ func testLogin(cmd *cobra.Command, httpClient *http.Client, server string, crede
 
 	return nil
 }
+
+// The client factory wraps the transport in a spinner, so the setting has to be
+// applied to the transport underneath rather than to the wrapper.
+func skipTlsVerification(roundTripper http.RoundTripper) {
```
NickJosevski: from my claude session:

> skipTlsVerification recurses past the spinner and assigns TLSClientConfig on that shared object, so unrelated clients stop verifying certificates.

pkg/cmd/login/login.go
```diff
@@ -131,7 +131,7 @@ func loginRun(cmd *cobra.Command, f factory.Factory, isPromptEnabled bool, ask q
 			httpClient.Transport = &http.Transport{}
 		}
 
-		httpClient.Transport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
+		skipTlsVerification(httpClient.Transport)
```
NickJosevski: from my claude session:, in general:

> http.Client{}, never passed to the function, flips from rejecting a self-signed cert to accepting it; cloning instead leaves it rejecting.

NickJosevski:
@flin-8 from the session I had with Claude from the start on this:

```
- NewSpinnerRoundTripper sets Next: http.DefaultTransport, the process-wide *http.Transport every client without its own uses.
- skipTlsVerification recurses past the spinner and assigns TLSClientConfig on that shared object, so unrelated clients stop verifying certificates.
- Measured: a plain &http.Client{}, never passed to the function, flips from rejecting a self-signed cert to accepting it; cloning instead leaves it rejecting.
- Replacing the whole config also clears NextProtos:[h2 http/1.1], the ALPN list that advertises HTTP/2.
- Not observable in the shipped CLI: client_factory.go:124 already sets InsecureSkipVerify on http.DefaultTransport for every command.
```

NickJosevski:
I'm not sure how I hit the close button 😊

NickJosevski (approved):
I'm ok for this to go, but just put in the comments I pulled from what I was doing with Claude in this space.

NickJosevski:
I threw that Claude session at this also yesterday, and it proposed this PR: 

https://github.com/OctopusDeploy/cli/pull/727

PR #727, opened 2026-09-09
fix: stop login --ignore-ssl-errors panicking on the spinner transport

NickJosevski:
`login --ignore-ssl-errors` asserted that the http client's transport was an *http.Transport so it could reach the TLS config. Since bafc85c5 the client factory always wraps the transport in a SpinnerRoundTripper, so the assertion panics for anyone with a server already configured:

    panic: interface conversion: http.RoundTripper is
    *apiclient.SpinnerRoundTripper, not *http.Transport

Before that commit the round-tripper was only installed in interactive mode, so a non-interactive run got a nil client, login substituted a plain &http.Client{}, and the assertion held by luck. Interactive runs of --ignore-ssl-errors were already panicking.

Walk the round-tripper chain instead of asserting on it. The real transport is replaced with an insecure clone rather than edited in place, since it is usually http.DefaultTransport and shared with every other client in the process, and an unrecognised transport now reports an error rather than quietly leaving verification on after the caller asked for the opposite.


NickJosevski:
referenced PR is simpler, and has Frank's context

